PR #6, opened 2020-02-20
Add the Observation class

ziotom78:
This PR implements a class `Observation`, which should provide a basic interface to TODs. It is meant to mimick the concept of «[observation](https://toast-cmb.readthedocs.io/en/latest/data.html)» in TOAST, so that it might eventually provide an high-level wrapper to it.

This PR is WIP; things to implement:

- [x] Definition of the class
- [x] Support for MJD times
- [x] Unit tests
- [X] Documentation

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -15,7 +15,6 @@ import sys
 
 sys.path.insert(0, os.path.abspath(os.path.join("..", "..")))
 
-
 # -- Project information -----------------------------------------------------
 
 project = "LiteBIRD simulation pipeline"
@@ -34,6 +33,8 @@ release = "0.1.0"
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.doctest",
+    "sphinx.ext.mathjax",
+    "sphinx.ext.napoleon",
 ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -12,6 +12,7 @@ Welcome to litebird_sim's documentation!
 
    tutorial
    simulations
+   observations
 
 Indices and tables
 ==================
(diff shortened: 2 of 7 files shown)


ziotom78:
The PR should be ready now; it just provide some basic functionality, but it might eventually become a wrapper around TOAST's class with the same name.

Please comment before **Monday 9 Mar, 12:00 UTC**; if I see no objections, I'll merge this.

mreineck:
I'm not sure whether `poetry.lock` should be in the repo. If not, maybe add it to `.gitignore`?

mreineck:
Also, perhaps we shouldn't say that an `Observation` is typically associated with a rather short time span. At some stages in the pipeline it will be more efficient to have long observations, for example during 4pi convolution. At later stages, where time-dependent detector characteristics play a role, it will most likely be necessary to split these long observations into smaller chunks. (And for simulations that are not primarily concerned with these time-dependent detector characteristics, it may actually be best to have one observation object for the entire simulation duration.)

ziotom78:
> I'm not sure whether `poetry.lock` should be in the repo. If not, maybe add it to `.gitignore`?

No, it's important that `poetry.lock` be included in the repository. The [Poetry manual](https://python-poetry.org/docs/cli/) says:

> If there is a poetry.lock file in the current directory, it will use the exact versions from there instead of resolving them. This ensures that everyone using the library will get the same versions of the dependencies.

Keeping it in the repository ensures that everybody who downloads the code will use the dependencies (e.g., NumPy, SciPy, etc.) pinned at the same version as everybody else.

> Also, perhaps we shouldn't say that an Observation is typically associated with a rather short time span. At some stages in the pipeline it will be more efficient to have long observations, for example during 4pi convolution. At later stages, where time-dependent detector characteristics play a role, it will most likely be necessary to split these long observations into smaller chunks. (And for simulations that are not primarily concerned with these time-dependent detector characteristics, it may actually be best to have one observation object for the entire simulation duration.)

I understand your point, I'll fix the sentence.

PR #21, opened 2020-03-23
Fix Windows configuration on appveyor

dpole:


ziotom78:
This was *really* hard! The original problem was the fact that the `pip` executable is not available in the same folder as `python` (it lives in the subfolder `scripts`); failing to set `PATH` properly made the shell call `pip-2.7` instead of the proper executable. Other problems were caused by the weird way used by `cmd` to pass the value of environment variable to child processes, and by the fact that `black` complains about the formatting of some files. I did not investigate the latter, as we can assume that the check on code formatting is already done by the Travis build under Linux (which should always be our reference point.)

PR #22, opened 2020-04-19
Add functions to distribute the workload among MPI processes

ziotom78:
This PR is a WIP. The aim is to implement a structure that permits to handle the distribution of the workload among MPI proceses. It tries to follow the same logic as TOAST, because it splits timelines into «observations»: in this way, integrating the code with TOAST should be (hopefully) as simple as possible.

Things to do:

- [X] Add a conditional dependency on `mpi4py`, so that the code can work with and without it
- [X] Add some functions to split a timeline between MPI processes
- [X] Modify the Travis CI build to test various MPI implementations under Linux and OSX (no MPI support for Windows is foreseen at the moment)
- [X] Add some documentation

### DIFF
--- a/litebird_sim/__init__.py
+++ b/litebird_sim/__init__.py
@@ -1,5 +1,7 @@
 # -*- encoding: utf-8 -*-
 
+from .distribute import distribute_evenly, distribute_optimally
+from .detectors import Detector
 from .healpix import (
     nside_to_npix,
     npix_to_nside,
@@ -9,9 +11,8 @@ from .healpix import (
     write_healpix_map_to_hdu,
     write_healpix_map_to_file,
 )
-from .distribute import distribute_evenly, distribute_optimally
+from .mpi import HAVE_MPI4PY, MPI_RANK, MPI_SIZE, MPI_COMM_WORLD, MPI_CONFIGURATION
 from .observations import Observation
-from .detectors import Detector
 from .simulations import Simulation
 
 __author__ = "The LiteBIRD simulation team"
@@ -28,6 +29,12 @@ __all__ = [
     # distribute.py
     "distribute_evenly",
     "distribute_optimally",
+    # mpi.py
+    "HAVE_MPI4PY",
+    "MPI_RANK",
+    "MPI_SIZE",
+    "MPI_CONFIGURATION",
+    "MPI_COMM_WORLD",
     # observations.py
     "Observation",
     # detectors.py
--- a/litebird_sim/__init__.py
+++ b/litebird_sim/__init__.py
@@ -11,7 +11,7 @@ from .healpix import (
     write_healpix_map_to_hdu,
     write_healpix_map_to_file,
 )
-from .mpi import MpiWrapper, MPI_COMM_WORLD
+from .mpi import MPI_COMM_WORLD, MPI_ENABLED, MPI_CONFIGURATION
 from .observations import Observation
 from .simulations import Simulation
 
@@ -30,8 +30,9 @@ __all__ = [
     "distribute_evenly",
     "distribute_optimally",
     # mpi.py
-    "MpiWrapper",
     "MPI_COMM_WORLD",
+    "MPI_ENABLED",
+    "MPI_CONFIGURATION",
     # observations.py
     "Observation",
     # detectors.py
--- a/litebird_sim/__init__.py
+++ b/litebird_sim/__init__.py
@@ -11,7 +11,7 @@ from .healpix import (
     write_healpix_map_to_hdu,
     write_healpix_map_to_file,
 )
-from .mpi import HAVE_MPI4PY, MPI_RANK, MPI_SIZE, MPI_COMM_WORLD, MPI_CONFIGURATION
+from .mpi import MpiWrapper, MPI_COMM_WORLD
 from .observations import Observation
 from .simulations import Simulation
 
@@ -30,10 +30,7 @@ __all__ = [
     "distribute_evenly",
     "distribute_optimally",
     # mpi.py
-    "HAVE_MPI4PY",
-    "MPI_RANK",
-    "MPI_SIZE",
-    "MPI_CONFIGURATION",
+    "MpiWrapper",
     "MPI_COMM_WORLD",
     # observations.py
     "Observation",
--- a/litebird_sim/__init__.py
+++ b/litebird_sim/__init__.py
@@ -9,12 +9,15 @@ from .healpix import (
     write_healpix_map_to_hdu,
     write_healpix_map_to_file,
 )
+from .distribute import distribute_evenly, distribute_optimally
 from .observations import Observation
+from .detectors import Detector
 from .simulations import Simulation
 
 __author__ = "The LiteBIRD simulation team"
 __version__ = "0.1.0"
 __all__ = [
+    # healpix.py
     "nside_to_npix",
     "npix_to_nside",
     "is_npix_ok",
@@ -22,6 +25,13 @@ __all__ = [
     "get_pixel_format",
     "write_healpix_map_to_hdu",
     "write_healpix_map_to_file",
+    # distribute.py
+    "distribute_evenly",
+    "distribute_optimally",
+    # observations.py
     "Observation",
+    # detectors.py
+    "Detector",
+    # simulations.py
     "Simulation",
 ]
--- a/litebird_sim/distribute.py
+++ b/litebird_sim/distribute.py
@@ -101,18 +101,8 @@ def _num_of_workers(arr, n, maxLen, weight_fn):
 
 
 def _find_max_and_sum(arr, weight_fn):
-    max_value = None
-    sum_value = 0
-    for cur_element in arr:
-        cur_weight = weight_fn(cur_element)
-        if not max_value:
-            max_value = cur_weight
-        elif max_value < cur_weight:
-            max_value = cur_weight
-
-        sum_value += cur_weight
-
-    return (max_value, sum_value)
+    weights = list(map(weight_fn, arr))
+    return (max(weights), sum(weights))
 
 
 def _partition(arr, n, k, weight_fn):
--- a/litebird_sim/distribute.py
+++ b/litebird_sim/distribute.py
@@ -22,11 +22,11 @@ def distribute_evenly(num_of_elements, num_of_groups):
 
     .. doctest::
 
-      >>> distribute_evenly(8, 3)
-      [(0, 3), (3, 3), (6, 2)]
+      >>> distribute_evenly(5, 2)
+      [Span(start_idx=0, num_of_elements=3), Span(start_idx=3, num_of_elements=2)]
 
-      >>> distribute_evenly(2, 3)
-      [(0, 1), (1, 1), (2, 0)]
+      >>> distribute_evenly(1, 2)
+      [Span(start_idx=0, num_of_elements=1), Span(start_idx=1, num_of_elements=0)]
 
     Args:
         num_of_elements (int): The number of elements to distribute
@@ -159,7 +159,7 @@ def distribute_optimally(elements, num_of_groups, weight_fn=None):
       >>> distribute_optimally([10, 10, 10, 10], 2)
       [Span(start_idx=0, num_of_elements=2), Span(start_idx=2, num_of_elements=2)]
 
-      >>> distribute_optimally([10, 10, 10, 10], 2)
+      >>> distribute_optimally([10, 10, 10, 20], 2)
       [Span(start_idx=0, num_of_elements=3), Span(start_idx=3, num_of_elements=1)]
 
       >>> distribute_optimally([40, 10, 10, 10], 2)
--- a/litebird_sim/distribute.py
+++ b/litebird_sim/distribute.py
@@ -1,0 +1,215 @@
+# -*- encoding: utf-8 -*-
+
+from collections import namedtuple
+
+Span = namedtuple("Span", ["start_idx", "num_of_elements"])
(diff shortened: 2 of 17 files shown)


ziotom78:
Ok, with the last commit I believe that this can be merged. Please comment!

dpole:
Hi Maurizio, sorry if my comments come late. I'd like to make a few proposals

litebird_sim/distribute.py
```diff
@@ -0,0 +1,215 @@
...
+        elements, each of them being a 2-element tuple containing (1)
+        the index of the element in the group and (2) the number of
+        elements in the group. Being named tuples, you can access the
+        index using the field name ``start_idx``, and the number of
+        elements using the name ``num_of_elements``.
+
+    """
+    assert num_of_elements > 0
+    assert num_of_groups > 0
+
```
dpole, line 46: It may look obscure, but what about
```
cur_pos = 0
res = []
for i in range(num_of_groups): 
    cur_length = (num_of_elements - (i + 1)) // num_of_groups + 1
    result.append(Span(start_idx=cur_pos, num_of_elements=cur_length))
    cur_pos += cur_length
````

File: litebird_sim/distribute.py [hunk 2]
```diff
@@ -0,0 +1,215 @@
...
+        if total > maxLen:
+            num_of_workers += 1
+            # Reset "total" for the next worker
+            total = cur_weight
+
+    return num_of_workers
+
+
+def _find_max_and_sum(arr, weight_fn):
```
dpole, line 103: what about
```
weights = list(map(weight_fn, arr))
return (max(weights), sum(weights))
```

litebird_sim/distribute.py
```diff
@@ -0,0 +1,215 @@
...
+from collections import namedtuple
+
+Span = namedtuple("Span", ["start_idx", "num_of_elements"])
+"""A sub-range in a sequence of elements.
+
+It has two fields: `start_idx` and `num_of_elements`.
+
+"""
+
+
```
dpole, line 12: I think that all the functions in this module can be avoided or simplified a lot by using `scipy.optimize.linear_sum_assignment`

litebird_sim/__init__.py
```diff
@@ -9,19 +11,34 @@
     write_healpix_map_to_hdu,
     write_healpix_map_to_file,
 )
+from .mpi import HAVE_MPI4PY, MPI_RANK, MPI_SIZE, MPI_COMM_WORLD, MPI_CONFIGURATION
 from .observations import Observation
 from .simulations import Simulation
 
 __author__ = "The LiteBIRD simulation team"
 __version__ = "0.1.0"
 __all__ = [
```
dpole, line 20: I'm not sure it is a good idea....but what do you think about delegating the definition of `__all__`  ot the modules and do simply `from module import *` in `__init__.py`?

litebird_sim/simulations.py
```diff
@@ -209,3 +233,66 @@ def flush(self):
...
+    def distribute_workload(self, observations: List[Observation]):
+        if self.mpi_size == 1:
+            self.observations = observations
+            return
+
+        cur_rank = self.mpi_rank
+        span = distribute_optimally(
+            elements=observations,
+            num_of_groups=self.mpi_size,
+            weight_fn=lambda obs: obs.nsamples,
```
dpole, line 293: Can you make an example where it is convenient to pass a weight functions instead of the weights directly?

litebird_sim/simulations.py
```diff
@@ -209,3 +233,66 @@ def flush(self):
             self.base_path / "report.html", "w", encoding="utf-8",
         ) as outf:
             outf.write(html_full_report)
+
+    def create_observations(
```
dpole, line 237: I'm not clear about the case of use of this method. The idea is that the Observation instances will create themselves later?

litebird_sim/mpi.py
```diff
@@ -0,0 +1,44 @@
+# -*- encoding: utf-8 -*-
+
+try:
+    import mpi4py
+    from mpi4py import MPI
```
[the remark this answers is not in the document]
ziotom78: Did you run it on a login node? This is a well-known limitation of NERSC login nodes, as explained here: https://docs.nersc.gov/programming/high-level-environments/python/mpi4py/

The fact that this error does not occur in Anaconda is not because this environment is «saner»; quite the contrary: it is unable to properly initialize the MPI system.

litebird_sim/mpi.py
```diff
@@ -0,0 +1,44 @@
...
+
+    #: Either a dictionary containing the mpi4py configuration, or
+    #: ``None`` if mpi4py is not available
+    MPI_CONFIGURATION = mpi4py.get_config()
+
+    #: Either the constant ``mpi4py.MPI.COMM_WORLD``, or ``None``
+    #: if mpi4py is not available
+    MPI_COMM_WORLD = MPI.COMM_WORLD
+
+except ModuleNotFoundError:
```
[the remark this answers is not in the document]
ziotom78: You're right, it's better to add the ability to pick MPI using an environment variable. I'll fix this.

litebird_sim/mpi.py
```diff
@@ -0,0 +1,44 @@
...
+    #: currently running, otherwise it's 1.
+    MPI_SIZE = 1
+
+    #: Either a dictionary containing the mpi4py configuration, or
+    #: ``None`` if mpi4py is not available
+    MPI_CONFIGURATION = None
+
+    #: Either the constant ``mpi4py.MPI.COMM_WORLD``, or ``None``
+    #: if mpi4py is not available
+    MPI_COMM_WORLD = None
```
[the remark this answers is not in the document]
ziotom78: I too had this idea, but then I discarded it because I feared that people would port their old MPI-based codes carelessly, without even checking if it works without MPI or not. But after your comment, I realized that this risk is non-existent: if we just provide `MPI_COMM_WORLD.rank` and `MPI_COMM_WORLD.size`, the code will break at the first instance of `sendv`, `recv`, `gather`, etc.

I'll fix this too.

litebird_sim/simulations.py
```diff
@@ -59,14 +53,44 @@ class Simulation:
...
+        if use_mpi and not HAVE_MPI4PY:
+            raise RuntimeError("use_mpi=True, but MPI is not available")
+
+        self.use_mpi = use_mpi and HAVE_MPI4PY
+        if use_mpi:
+            self.mpi_rank = MPI_RANK
+            self.mpi_size = MPI_SIZE
+        else:
+            self.mpi_rank = 0
+            self.mpi_size = 1
```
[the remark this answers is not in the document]
ziotom78: Right.

litebird_sim/distribute.py
ziotom78: Is this to avoid the `if` within the `for` loop? Do I understand correctly?

litebird_sim/distribute.py
ziotom78: Right, my first implementation used Numba and I decided to unroll the `for` loop to avoid iterating the same list twice. However, I realized that this function is not worth to be optimized, as it has to process very few elements. I'll apply your definition.

litebird_sim/distribute.py
ziotom78: Mmm… I have never used that function, and it's difficult for me to understand how it works and what's the meaning of the return values. Can you suggest a patch that uses it? Ideally, you should put comments that explain these things, because I expect that other people might have the same problems as me in understanding it.

litebird_sim/__init__.py
ziotom78: From what I understand, `__all__` is used by Python whenever it finds the statement `from <library> import *`, and by code linters that check imports. As described in this [Stack Overflow question](https://stackoverflow.com/questions/31079047/python-pep8-class-in-init-imported-but-not-used), you really need to put `__all__` in the module, not in the sub-modules.

litebird_sim/simulations.py
ziotom78: The example you highlighted (lines 290-293) is the perfect example: we pass a list of `Observation` objects, where the weight is not the object itself but rather its `nsamples` field. This line is the reason why I added the `weight_fn` parameter to the functions in this module.

litebird_sim/simulations.py
ziotom78: No, the method `create_observations` initializes a sequence of `Observation` objects (line 264). This is the method that will be called in virtually *any* pipeline written using our `litebird_sim` framework, following this logic:

```python
sim = lbs.Simulation()
detectors = [lbs.Detector(params) for params in …]
lbs.create_observations(detectors, …)
# Etc.
```

litebird_sim/mpi.py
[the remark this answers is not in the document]
ziotom78: Done.

litebird_sim/mpi.py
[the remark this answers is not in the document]
ziotom78: Done.

litebird_sim/distribute.py
ziotom78: Done.

litebird_sim/distribute.py
dpole: Sure, I'll propose a single function for the two functions in this module

litebird_sim/distribute.py
dpole: Yes, the if and also this mechanism of the leftovers. But don't bother for now, let's discuss on the patch I'll propose maybe later today

dpole:
I am not clear about how the user should decide to run parallel or serial code.
I thought that the user decides at runtime (and the code crashes if he tried to use MPI when he couldn't).
I imagine that, when he wants to run serial code, he only wants to get a `comm` object with the `size` and `rank` attributes (respectively set to 1 and 0), the code that will receive this object will know that when size = 1 it has to run serially.
When the user wants MPI instead, he expect comm to be a genuine MPI communicator.
In this mindset, `mpi.py` should contain only

```
def get_comm(mpi=False):
    if mpi:
        from mpi4py.MPI import COMM_WORLD
        return COMM_WORLD
    else:
        return SerialWorldComm()

class SerialWorldComm:
    rank = 0
    size = 1
```
What do you and others think?

ziotom78:
> I am not clear about how the user should decide to run parallel or serial code.

I see three cases:

1.  The "user" (i.e., anybody developing a pipeline that uses `litebird_sim`) doesn't care about MPI. In this case, they just ignore `MPI_COMM_WORLD`;
2.  The user wants their code to be runnable both serially and using MPI; it chooses what is the case using the environment variable `LITEBIRD_SIM_MPI=1` or `LITEBIRD_SIM_MPI=0`
3.  The user has developed some code that works under MPI, and they do not want/can support serial communication because it's too complex. Then they pass `use_mpi=True` to `Simulation.__init__`.

ziotom78:
@dpole, any other comments on this PR? I would like to merge it soon and start working on something else.

litebird_sim/distribute.py
dpole: I realized that function doesn't suit our case of use perfectly. Fine with your approach!

litebird_sim/simulations.py
dpole: What `_` would represent here? I'm not clear if this stage should allocate the TOD etc or it is just creating a container for the observation setup and metadata

dpole:
Thanks for the reminder: this felt off my radar. About MPI, I think I'm still confused. If the user has to provide the information about whether to use MPI or not through a keyword argument, he can pass directly the communicator or `None` -- where `None` means don't use MPI). Even in this case, I'm missing the role of the `MpiWrapper` class in `mpi.py`.

litebird_sim/simulations.py
ziotom78: In this stage the time span of the whole simulation is split among detectors. The real TOD must be created by the scanning strategy simulator; the function is just creating the «containers», as you call it. It's at this stage that the code slices the timespan figures out which MPI process will process each of them.

dpole: Ok, thanks. I was wondering whether every detector should store quantities that are shared by all of them, like the sampling frequency

ziotom78:
> Even in this case, I'm missing the role of the `MpiWrapper` class in mpi.py.

That's a valid concern. My idea was to create some middleware that could be extended to monitor MPI calls (e.g., to track how much data is exchanged between processes), with the purpose to include this information in the report output by the code. But at the moment the class is more of an empty shrine.

litebird_sim/simulations.py
ziotom78: Mmm… At the moment the sampling frequency is the same among the detectors, but I would not bet it will stay as it is forever. But I agree, there are some types of information that would be worth to be shared by many detectors, like the quaternion representing the rotation of the spacecraft. This does not fit in the `Observation` class, so it should perhaps be part of the `Simulation` object.

dpole:
This functionality is likely to be needed only on computational centers like nersc. They typically have a set of top notch profilers that I would favor for this kind of purpose. What do you think?

ziotom78:
> This functionality is likely to be needed only on computational centers like nersc. They typically have a set of top notch profilers that I would favor for this kind of purpose. What do you think?

I have found this feature very useful when debugging codes on my own laptop, but now I realize that  I've had done this thing so many times that it's probably worth to move `MpiWrapper` on a package on its own. I'll fix this PR as soon as I have some time.

ziotom78:
Ok, I think I have addressed all the comments. If no further objections arise, I would like to merge this PR in the weekend.

PR #30, opened 2020-05-19
Add healpy 1.13 as a dependency

ziotom78:
After having merged [PR#29](https://github.com/litebird/litebird_sim/pull/29), it should be possible to import Healpy without breaking the Appveyor script. This is fundamental to start implementing the pointing-generation module.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -17,7 +17,7 @@ version = "1.4.3"
 [[package]]
 category = "main"
 description = "Disable App Nap on OS X 10.9"
-marker = "python_version >= \"3.3\" and sys_platform == \"darwin\" or platform_system == \"Darwin\""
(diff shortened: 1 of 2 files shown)


ziotom78:
All tests have passed, so I'm merging this now.

mreineck:
Just curious: are there already concrete plans for the pointing-generation module? Depending on which `healpy` functionality it requires, it might be possible to interface it with `pyHealpix`, which is probably easier to install.

ziotom78:
> Just curious: are there already concrete plans for the pointing-generation module?

I plan to implement a very naive module which just composes a few quaternions, waiting for TOAST to be `pip`-installable to integrate its pointing-generation code (if we think it's suitable).

Using pyHealpix would be great, as I remember that it's much faster than Healpy. But how should we declare this dependency? Is it already available in PyPI?

PR #32, opened 2020-05-28
add cxxbase package

mreineck:
This is a first suggestion how the "cxxbase" code could be added to litebird_sim. Directory names, package naming etc. are completely free to discuss; I'm happy about any suggestions.

I dont know how to integrate the package's `setup.py` with the central `poetry` mechanism, so I need help there.


ziotom78:
> I dont know how to integrate the package's setup.py with the central poetry mechanism, so I need help there.

I checked, and what's needed is to add the following lines to `pyproject.toml`:

```toml
[tool.poetry]
build = "build.py"
```

Then you need to write a `build.py` file. The gist here shows how to compile Cython and pybind11 modules:

https://gist.github.com/dmontagu/b91d5fd5319ae6fe004ecd28771d985e

Reference: see https://github.com/python-poetry/poetry/issues/1351#issuecomment-527775913

mreineck:
Hmm, I'm using neither `cython` nor `cmake`. I'll see how I can adjust this.

tskisner:
Will there be any future conflicts if we incorporate `cxxbase` components into `healpy` and then import it in a litebird_sim module?  Imagine this scenario:

1. healpy compiled extension links to `libcxxbase` (or whatever, which is bundled with healpy.
2. a litebird_sim module imports `pyHealpix`, which also links to "libcxxbase".

The dynamic loader will only open `libcxxbase` once, so the order of imports will determine which version of the library gets loaded.

We see this same problem in toast if we make a compiled extension that uses MKL.  If you also use a version of numpy that uses MKL, then importing numpy and toast in different orders changes which MKL is loaded and used by **both** packages.  This can cause segfaults or random corruption unless both packages were originally built with the same upstream library version.

Some options would be:

1.  Have cxxbase as an independent package available on PyPI with the python bindings (which is a large support cost, I know).
2.  Have at least one of the packages link to cxxbase statically.

I guess it is probably easier for healpy to do (2).  But I just wanted to raise this issue.

mreineck:
I guess the trick would be to include version numbers automatically into the package (and library) names. But I'll need to do some more reading before I know how to do this ...

mreineck:
So this `build.py` should actually sit in the root directory? OK, then things start to make a bit more sense :)

dpole:
Sorry for jumping in late. This is what I wanted to warn about in the other thread. If a package has a non trivial install process, this process has to be merged with the build of the main repo. If we have many modules like that, maybe we'll get to the point where we have conflicts in some setups that the modules have to share.

The other possibility is to keep your repo at the organization level. This would keep a more modular install process. We can still add a module to litebird_sim that just contain the bindings to your lib. The user wouldn't notice the difference

mreineck:
Having a separate repo at the organization level is perfectly fine for me as well!

ziotom78:
Yes, having read @dpole 's concerns, I agree that having complex C++ modules kept in another repo is more elegant. Regarding poetry, after we have created `cxxbase`, I believe that just running

    poetry add git+https://github.com/litebird/cxxbase

should be enough.

@mreineck , are you able to create repositories in the `litebird` organization?

mreineck:
> @mreineck , are you able to create repositories in the litebird organization?

Doesn't look like it. When I'm in my own organization, I see a "New" button, but in litebird_sim I don't have it.

dpole:
Let me fix it

dpole:
The problem was that you are a member of the repo but not a member of the organization. You should have received an invitation

mreineck:
Thanks! That seems to work.

dpole:
One thing, Martin, from the outside I don't find the cxxbase name very descriptive. Are there other ideas on the table? :)

dpole:
I just noticed that the content is very heterogeneous, now I understand why the name was so generic

mreineck:
> One thing, Martin, from the outside I don't find the cxxbase name very descriptive. Are there other ideas on the table? :)

You are absolutely right! I have been thinking about a better name, but whatever we choose, I fear it will remain pretty nonspecific, as the package does quite a few things which are only loosely related. It contains:

- a pretty powerful FFT library (which has been adopted by `numpy` and `scipy`)
- the spherical harmonic transforms from `libsharp` (which uses the FFT)
- code for non-equidistant Fourier transforms
- the 4pi convolution/interpolation package (which uses all of the above)
- basic Healpix functionality (`ang2pix` and friends, `query_disc` and friends)
- a gridder/degridder for radio astronomy (using FFTs and non-equidistant FTs), which is now also used in `wsclean`

I want to distribute all of this together because the components share a huge part of common code. Of course I could separate out all of the common code in a single C++ library from which all other packages depend, but then installation on the user side becomes much more complicated.

If anyone has a good name suggestion for this, I'd be happy to hear it! My best attempt so far is `DUCC` ("d...efinitely useful code collection"), but it may be possible to improve on that!

mreineck:
@tskisner:

> Will there be any future conflicts if we incorporate cxxbase components into healpy and then import it in a litebird_sim module?

How about this: the whole package will generate a single shared library (at the moment it doesn't, but I think I know how to achieve this) called `cxxbase1`. Once I make a significant change, I'll bump this to `cxxbase2` etc. The names of the shared libraries will be distinct, and in user code one could do something like `import cxxbase2 as cxxbase`.

What do you think?

mreineck:
Closing this, but we can continue the name discussion here of course!

dpole:
> > One thing, Martin, from the outside I don't find the cxxbase name very descriptive. Are there other ideas on the table? :)
> 
> You are absolutely right! I have been thinking about a better name, but whatever we choose, I fear it will remain pretty nonspecific, as the package does quite a few things which are only loosely related.

Yes, without splitting the package into subpackages only very generic names come to my mind, like cxxsky, cxxskytools or similar

mreineck:
Waaay too specific ;-)

SHTs are being used in geology, electron microscopy and many other areas, non-equidustant FTs are used predominantly in computer tomography, and FFTs are of course used everywhere ...

I'm of course planning to offer the individual sections via subpackages (similar to `scipy`), so we can get more specific at this level!

Issue #35, opened 2020-06-15
Detectors as indices of arrays instead of objects


dpole:
It would be extremely good (and timely) to take an explicit decision about the data layout.

### Current structure
Every simulation owns a list of observations. Every observation owns a detector from which you can get a timestream array, a quaternion, a PSD array, a white noise level etc. Firstly, I would reduce the number of indirections from `simulation.observations[3].detector.get_tod()` to `simulation.observations[3].tod`. Anyway, even in the latter case, the paradigm is
 
```
sim.observations[3].tod  # 1D np.array
sim.observations[3].wn_level  # float
print(sim.observations[3])
```

The advantage is that the code reflects the logical hierarchy and organization of the data we are storing. On the practical side, it allows to easily plug in and remove detectors, print or dump to disk all the attributes of the detector. This model is well suited for random operations on random detectors.

### Proposal
I propose the following alternative paradigm. The simulation directly owns an array of quaternions, a 2d array of TODs, a 2d array of PSDs, a 1d array of white noise levels etc. These arrays stack the data of the individual detectors.

```
sim.tod  # 2D np.ndarray: first index for detectors, second for time
sim.tod[3]  # same result as before
sim.wn_level[3]  # same result as before
sim.print_detector(3)  # same result as before
```

### Pros
Operations that involve more than one detector are likely to be easier to write. This is sure if the operation involves all the detectors.
If these operations are done by compiled codes, the interface is much easier (pass one big array, instead of many small values/arrays)
All numpy functions integrate seamlessly with `ndarray`s
...many other technical/philosophical things that probably shouldn’t drive the decision...

### Examples
Standard deviations of the TOD of each detector: `sim.tod.std(1)`
TOD of the standard deviation across the focal plane: `sim.tod.std(0)`
TOD of the mean signal across the focal plane: `sim.tod.mean(0)`
Add white noise to the TOD: 
`sim.tod += np.random.normal(size=sim.tod.shape) * sim.wn_level[:, np.newaxis]`
Add cross-talks given by a crosstalk matrix `C`: `sim.tod = C @ sim.tod`

### Note
If we think it is necessary, we could still implement our current API with a caveat: `sim.observations[3].tod` has to be a view of `sim.tod[3]` (same for other attributes). In practice, it means that `sim.observations[3].tod = …` is not allowed (you have to do `sim.observations[3].tod[:] = …`) and `sim.observations[3].wn_level` is a 1-element array.

[...]

PR #42, opened 2020-06-30
Data layout proposal

dpole:
This PR proposes an implementation of the data layout proposed in #35.

# Summary
* an `Observation` contains multiple detectors
* information about the detectors
  - become an attribute of the `Observation`, if shared by the detectors
  - are stacked in an array, which become an attribute of the observation. Notably, `obs.tod` is a 2d array with shape `(n_det, n_sample)`
* `Detector` is not needed anymore

# Motivation
Stacking data that is likely to interact into regular data-structures typically allows easier APIs (and sometimes faster computation). E.g. `obs.tod.mean(0)` computes the common mode; `C.dot(obs.tod)` applies the cross-talk matrix `C` using threaded and highly optimized math libraries like MKL.

# Perspective
Prefer shared-memory parallelism over MPI parallelism: the former typically have no or minimal API overhead compared to serial code (and sometimes better performance compared to in-node MPI parallelization). In this perspective, it is better to have as much data as possible packed in a compact data structure.

# Next steps (maps data)
In this perspective, operations on an `Observation` should avoid MPI parallelization and therefore TOD-to-map accumulation should be a serial operation (maps are not distributed, every MPI rank holds an entire map). This should be possible as LiteBIRD maps should take only a limited fraction of any foreseen node size


dpole:
@ziotom78 I think I need a feedback to make sure I'm not breaking your pieces of code -- in addition to a general feedback and on the model

litebird_sim/simulations.py
```diff
@@ -322,52 +322,46 @@ def flush(self):
...
-                    use_mjd=use_mjd,
-                )
-                observations.append(cur_obs)
-                span_s = cur_sampfreq_hz * nsamples
-
-                if use_mjd:
-                    cur_time = Time(mjd=cur_time) + TimeDelta(seconds=span_s)
-                else:
-                    cur_time += span_s
+                cur_start_time += span_s
```
ziotom78, line 349: This line of code fails, because `cur_start_time` was never initialized above.

litebird_sim/simulations.py
```diff
@@ -322,52 +322,46 @@ def flush(self):
...
             else:
-                if use_mjd:
-                    cur_time = start_time
-                else:
-                    cur_time = start_time.mjd
-
-            for cur_obs_idx in range(num_of_obs_per_detector):
-                nsamples = samples_per_obs[cur_obs_idx].num_of_elements
-                cur_obs = Observation(
-                    detector=cur_det,
```
ziotom78, line 351: With your modifications, the parameter `detector` in `Observation.__init__` changed into `detectors`

litebird_sim/simulations.py
```diff
@@ -322,52 +322,46 @@ def flush(self):
...
+        self.observations = []
+        num_of_samples = sampfreq_hz * duration_s
+        samples_per_obss = distribute_evenly(num_of_samples, num_of_obs_per_detector)
+        start_samples = np.cumsum([0]+samples_per_obss[:-1])
+
+        # Keep only one every comm.size starting at my rank
+        start_samples = start_samples[self.mpi_comm.rank::self.mpi_comm.size]
+        samples_per_obss = samples_per_obss[self.mpi_comm.rank::self.mpi_comm.size]
+
+        for start_sample, samples_per_obs in zip(start_samples, samples_per_obss):
```
ziotom78, line 344: Beware that `samples_per_obss` is a list of `Span` objects, not of numbers.

litebird_sim/simulations.py
```diff
@@ -322,52 +322,46 @@ def flush(self):
...
-                    cur_time = Time(mjd=cur_time) + TimeDelta(seconds=span_s)
-                else:
-                    cur_time += span_s
+                cur_start_time += span_s
+
+            cur_obs = Observation(
+                detector=detectors,
+                start_time=cur_start_time,
+                sampfreq_hz=sampfreq_hz,
+                nsamples=samples_per_obs,
```
ziotom78, line 355: Since `samples_per_obs` is an element in a list of `Span` objects, it means that you are passing a `Span` object to the parameter `nsamples` instead of a scalar. This causes an error in `Observation.__init__` (line 65):

```python
self.tod = np.empty(self.detectors.shape+(nsamples,), dtype=dtype)
# TypeError: 'Span' object cannot be interpreted as an integer
```

ziotom78 (dismissed):
Hi @dpole , sorry for the delay in reviewing this. There are a few problems in the code, I have tried to be as clear as possible in detailing what's probably causing them.

mreineck:
From your presentation at the telecon I got the impression that you expect TOD to be processed in chunks of several hours or so. Maybe it's not really a concern, but this is somewhat at odds with the way that TOD are simulated from 4pi convolved data cubes.

The 4pi convolver has the following aspects:
1) it inputs the sky and beam a_lm
2) from those it generates the interpolation data cube (fairly expensive operation). The size of the data cube is 10-100 times the size of a sky map
3) given a set of (theta, phi, psi) pointings, it computes the TOD values; this operation is quick and scales with the length of the TOD. This can be called repeatedly with different pointing arrays.

As you see, there is a fairly large starting overhead and memory cost associated with this kind of TOD generation. It will definitely not be economical to setup and destroy the interpolator object for a few hours' worth of TOD.
On the other hand, given the size of the data cubes, it will not be possible either to hold all data cubes for all detectors in the memory of a single node. So, it will not be possible to do a full focal plane simulation on a single node, no matter how finely the TOD is chunked.

mreineck:
Sorry for not raising this point earlier ... I only realized the intended simulation strategy when I heard "we can fit several hours of full focal plane TOD and hundreds of maps onto a single compute node".

dpole:
Hi Martin, no problem. The important think is that we caught this point before a decision. I think that your application isn't necessarily in contrast with the current setup.

Just to clarify one thing: nothing prevents a specific application from having observations with a single detector in the TOD. The constrain that this PR adds is: 

* if many detectors are in the same node, you can assume that they are organized in a 2d array
* codes that involve communications between detectors can assume shared memory (off-node communications between detectors or time chunks will be rare)

Even if we decide not to go for this, we have to be aware that we are implicitly agreeing that any interaction between detectors has to pass through one (or many) MPI call.

Of course if all of our application involve communications across the full-mission time (like your case) it is better to assume that a processor holds the full tod of a mission. However, I think that interactions between detectors' tod will be frequent, and therefore it makes sense to strive for the data model I propose.

Accommodating cases that work over the full tod requires a bit of a stretch. In my mind tods arrays are blocks of a block-cyclic distributed matrix that can be reshuffled at runtime (even though it should be avoided, whenever possible). 

What do you think?

litebird_sim/observations.py
```diff
@@ -61,9 +61,15 @@ def __init__(self, detector, start_time, sampfreq_hz, nsamples, use_mjd=False):
...
             self.start_time = start_time
 
-        self.sampfreq_hz = sampfreq_hz
-        self.nsamples = nsamples
-        self.tod = None
+        self.detectors = np.array(detectors)
+        self.tod = np.empty(self.detectors.shape+(nsamples,), dtype=dtype)
+        for field in fields:
+            field_arr = np.array([some_method(field, d) for d in detectors])#XXX
+            setattr(self, field, field_arr)
```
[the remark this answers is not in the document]
dpole: The idea behind this for loop is that the observation loads from the IMO all the fields that will be relevant for the pipeline. For example: white noise level, position in the focal plane, beam...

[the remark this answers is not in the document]
dpole: It's not meant to be a working code, just a prototype of what I'd expect the workflow to be

litebird_sim/__init__.py
```diff
@@ -1,7 +1,6 @@
 # -*- encoding: utf-8 -*-
 
 from .distribute import distribute_evenly, distribute_optimally
-from .detectors import Detector
```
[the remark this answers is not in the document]
dpole: Yes, thanks for catching this. I forgot to plug this line back in

litebird_sim/observations.py
```diff
@@ -40,10 +40,10 @@ class Observation:
 
     """
 
-    def __init__(self, detector, start_time, sampfreq_hz, nsamples, use_mjd=False):
-        self.detector = detector
+    def __init__(self, detectors, start_time, sampfreq_hz, nsamples,
+                 use_mjd=False, fields=[], dtype=np.float32):
```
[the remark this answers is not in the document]
dpole: As mentioned above, in my idea it contains the information about all the quantities that have to be loaded, which depend on the application and therefore are left as a free parameter

test/mpi/test_mpi.py
```diff
@@ -4,6 +4,7 @@
 
 
 def test_distribution():
+    return #XXX
```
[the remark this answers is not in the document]
dpole: This test has to be rewritten (tests have to be written in general). The line is not supposed to stay there, it just cuts temporarily the test out

litebird_sim/simulations.py
dpole: Thanks for pointing this out. I've got to revisit a couple of things then

mreineck:
It's hard for me to judge how simulations will be carried out for LiteBIRD (e.g. how many detectors will be processed simultaneously, how long the chunks of simulated TOD are, etc.).
What I can say is the following: if we assume that the "source" of the TOD in a pipeline is the 4pi convolver, and if we ignore complicated MPI communications for a moment, then it will not be possible to process more than, say, 100 detectors on a single node, even if that node is quite big. There is no real way around this: if you want to be able to ask "give me the TOD of this detector in the time range from A to B", then memory of roughly 1GB per detector is required on the node to compute this information, independent of the time range.
It's of course always possible to generate the TOD and write them to disk in a separate run; then they can be loaded in chunks as needed. But I think this is not the desired application scenario, since the data volumes will be quite big and I/O can be a bottleneck.

dpole:
Ok, 1GB/detector is the requirement to keep in mind. I think that it is now clear why  4pi convolution prefers to have the entire tod of a detector in the node memory. 

I think (and please anybody feel free to object) that this is not incompatible with this PR. The proposal is not to force chunking tods in time such that you have all the detectors in a single processor. The proposal is more "if you do have more detectors in the local memory, their tod should be aligned in memory".

As I said before, if all the application will be like the 4 pi convolution, it isn't worth bothering with this requirement. However, my impression is that it is not the case: only few applications require a full mission tod in memory and the distribution of detectors across nodes. 

My understanding is that only beam studies require 4pi convolution. Most of the other studies can assume symmetric beams. So glitches, calibration drifts, crosstalks, etc. don't require extremely long tod chunks to be in the local memory. I believe that the best way to go is to implement this data model and provide a utilities to resize the chunks along the time and detector axes.

This adds some complication, but still I think it is less complicated than the alternative (i.e. requiring the time axis to be a very long, potentially a single, chunk and each line of the detector axis to be distributed), which implies that any user attempting operations that involve multiple detectors have to use an MPI interface.

mreineck:
> Ok, 1GB/detector is the requirement to keep in mind.

To be more specific, the rough formula is
`lmax**2 * (2*beam_mmax+1) * 60` bytes
for maximum multipole moment `lmax` and maximum azimuthal beam moment `beam_mmax`.

Depending on the required resolution, this can vary quite a bit, and if one doesn't need high resolution, fairly many convolution data cubes will fit onto a node.

I agree that this scenario is not incompatible with the PR.

litebird_sim/simulations.py
```diff
@@ -322,7 +322,7 @@ def flush(self):
 
     def create_observations(
         self,
-        detectors: list,
+        detectors: List[str],
```
ziotom78, line 325: Good to have the type specified. I was wondering if it would be better to call this parameter `detector_names` to make it more explicit (the same for the `Observation` class): now that there is a `Detector` class, people might get confused and think they're supposed to pass a list of `Detector` objects here.

dpole: I think that the type hint makes it already clear enough.

ziotom78: I agree it's clearer, but when I wrote my comment this morning I've just realized that type hints are not visible when reading code like this:

```python
import litebird_sim as lbs

detectors = some_complicated_function()
sim = lbs.Simulation()
# Wait, is "detectors" a list of strings or of some other type?
sim.create_observation(detectors=detectors, ...)
```

This looks clearer:

```python
import litebird_sim as lbs

detector_names = some_complicated_function()
sim = lbs.Simulation()
sim.create_observation(detector_names=detector_names, ...)
```

Moreover, I expect that other functions/modules/classes will expect a list of `Detector` objects as inputs (e.g., to generate noise). If we use the same keyword `detectors=` for these two cases, this will make the code less clear:

```python
# This expects a list of strings…
sim.create_observation(detectors=detectors, ...)
# …but this wants a list of "Detector" object
generate_noise(detectors=detectors, ...)
```

I think that our code should use the following conventions:

- The keyword `detector_names` indicates that strings are expected;
- The keyword `detectors` indicates that `Detector` objects are expected.

This would imply we need to fix my implementation of `Observation.__init__` as well, but I think it's worth it.

dpole: This is probably something that is more efficient to discuss at the call tomorrow. Very briefly, I think we should avoid passing Detectors, both for clarity and performance.

About clarity, I prefer if we pass only the arguments that the function needs instead of a Detector. This way the user knows exactly what is needed and can experiment with the function more easily (i.e. without creating a Detector object, which may be overcomplicated for the purpose). In general, I think we should strive for using as few new classes as possible: If a class can be reduced to a standard container, we should avoid defining new classes. Every new definition raises the barrier to the new users. In particular, it seems to me that we could use a simple dictionary (or SimpleNamespace) instead of a Detector.

About performance, all the examples that come to my mind run faster if the the arguments are passed with one array for each Detector attribute rather than a list of Detectors (or dictionaries). At least this is the idea behind this pull request.

I think that noise generation fits well both these arguments. I'd prefer the pattern to be

```
sim.create_observation(detectors=detectors, ...)
for obs in sim.observations:
    obs.tod += generate_noise(obs.fknee, obs.white_noise, ...)
```

giuspugl:
> My understanding is that only beam studies require 4pi convolution. Most of the other studies can assume symmetric beams. So glitches, calibration drifts, crosstalks, etc. don't require extremely long tod chunks to be in the local memory. I believe that the best way to go is to implement this data model and provide a utilities to resize the chunks along the time and detector axes.

What if we want to simulate all the systematics effects in one single simulation, i.e. combine beam asymmetries , HWP imperfections, bandpass mismatches, cosmic rays, etc.. ? My understanding is that this is something very needed for Litebird...  I would opt more for a data layout which could  account for this option too..

litebird_sim/simulations.py
ziotom78: > This is probably something that is more efficient to discuss at the call tomorrow.

Given that not many people are going to attend, I would prefer to continue the discussion here, so that others can chime in if they want.


> About performance, all the examples that come to my mind run faster if the the arguments are passed with one array

You're right, I discovered that creating 100,000  detectors using one list of Detectors takes 90 ms, while creating separate NumPy arrays only takes 50 ms. However, I do not expect that users are going to notice if the code spends 0,04 seconds more in running a function that's only called once in the lifetime of a program… (Moreover, LiteBIRD is going to have far less than 100,000 detectors!)

> I prefer if we pass only the arguments that the function needs instead of a Detector.

As I already said somewhere else, I do like this idea of yours that the framework should provide functions with the explicit list of parameters. In fact, my proposal idea is compatible with this code of yours:

```python
sim.create_observation(detectors=detectors, ...)
for obs in sim.observations:
    obs.tod += generate_noise(obs.fknee, obs.white_noise, ...)
```

but I mistakenly wrote my example above forgetting `sim.` in front of `generate_noise` (I fixed my comment):

```python
# This expects a list of strings…
sim.create_observation(detectors=detectors, ...)
# …but this wants a list of "Detector" object
sim.generate_noise(detectors=detectors, ...) # Add "sim."
```

The idea is that there is a function, `generate_noise`, that takes the arguments `fknee_hz`, `white_noise_ksqrthz`, etc., exactly like in your example, plus a `Simulation.generate_noise` method that wraps `generate_noise` and uses the IMO to retrieve these parameters coherently. In this way, we have the best of the two worlds. (I am following the same approach in PR #48 : `Simulation.generate_pointing_information` is a high-level wrapper around `Observation.generate_pointing_information` that knows how to access the IMO to retrieve the scanning strategy parameters.)

>  In particular, it seems to me that we could use a simple dictionary (or SimpleNamespace) instead of a Detector.

My initial implementation of `Detector` was a simple named tuple, but I changed it to a class because in this way we can provide implementations for `Detector.__repr__`, `Detector.__str__`, etc. I agree that this is not as awesome as in Julia, where you can define even a `plot` function over your custom types (see [here](https://ziotom78.github.io/Harlequin.jl/stable/genpointings/) for an example about scanning strategies, and [here](https://lspestrip.github.io/Stripeline.jl/latest/instrumentdb/#Introduction-1) for bandpasses), but this can still be helpful when debugging codes or doing exploratory work..

Moreover, named tuples are immutable, and this might be too limiting. The best option for this kind of classes would be to use [data classes](https://docs.python.org/3/library/dataclasses.html#module-dataclasses), but they are not supported in Python 3.6. Regarding dictionaries, I strongly dislke them, because it's too easy for people to abuse them with code like:

    detector["noise_{}".format(measurement_unit)] = ...

I want to know exactly what are the names of the fields in a compound object, and this kind of code prevents me from doing so by just reading at it.

There are surely disadvantages in using a `Detector` class (more code to write, more verboseness, more classes to document…), but let's list some advantages as well:

-   If one must create a vector `fknee_hz`, a vector `white_noise_hz`, a vector of quaternions, etc., there is the increased chance that one of these vectors will have `N + 1` elements instead of `N` (or `N - 1`, or… you get the idea), expecially if they are initialized at different places in the code. Another bug, far more difficult to catch, is that all the lists have the same number of elements, but in a different order! This kind of bugs is impossible if you use a list of `Detector` objects. **This is by far my strongest point in favour of the adoption of a `Detector` class.**

    I learned this lesson the hard way several years ago, when I wrote a C code that passed arrays of different lengths to the parameters `ttype`, `tform`, and `tunit` in CFITSIO [`fits_create_tbl`](https://heasarc.gsfc.nasa.gov/docs/software/fitsio/c/c_user/node43.html); I kept getting `SIGSEGV` crashes, until I found that I missed one element in one of the three parameters. Now I'm a strong advocate of arrays of data structures for this kind of situations, even if this means that the code will take 0,04 s more to complete.

-   For objects that require a complex initialization, using a dedicated class can lead to improved clarity and performance. For instance, an object might want to initialize derived quantities from the parameters passed to its constructor, if these derived quantities are going to be used several times during the lifetime of the object and/or they are expensive to compute. For instance, unlike Healpy, Healpix.jl uses a [Resolution](https://github.com/ziotom78/Healpix.jl/blob/ebc24731c0275207e454af898e946d28864feddf/src/resolution.jl#L24) object instead of a plain `NSIDE` parameter to specify the resolution for a map, because it caches a number of quantities derived from `NSIDE` that should otherwise be computed over and over again. The class' constructor is the most obvious place where to place this kind of initialization code.

-   If some people are going to extend an existing code by plugging in some more modules, there is the chance that the code already has the inputs they require. Example: you want to include noise generation in some code that performs beam convolution simulations. If you're loading detectors from the IMO, the noise parameters are already in place in the `Detector` objects, and you just have to call `sim.generate_noise(detectors=detectors)`. The opposite case means that you must figure out how to retrieve this information, and you must be able to **construct an array of values that has the same order as the other arrays (e.g., detector quaternions) used elsewhere in the code** (see the first point above in the list).

-   It's true that creating a `Detector` instance is verbose, because the constructor expects many fields. However, this might be mitigated by specifying `None` as the default value for all its fields but `name` (this is something I already wanted to do). This is going to be useful in unit tests, where you typically want to use only a few parameters for a `Detector` object, because you could just write:

    ```python
    def test_noise():
        ...
        det = Detector(name="a", fknee_hz=1.0, wn_level_ksqrthz=10.0) # Forget about other parameters
        sim.generate_noise(det)
        # Now check that the noise looks good
    ```

    And, again, if you're just going to directly call functions like `generate_noise`, you do not need to create a `Detector` class at all:

    ```python
    obs = Observation(...)
    arr = lbs.generate_noise(fknee_hz=1.0, wn_level_ksqrthz=10.0, nsamples=obs.nsamples)
    ```

ziotom78:
> What if we want to simulate all the systematics effects in one single simulation, i.e. combine beam asymmetries , HWP imperfections, bandpass mismatches, cosmic rays, etc.. ? My understanding is that this is something very needed for Litebird... I would opt more for a data layout which could account for this option too..

True, we have the task to provide the code to run end-to-end simulations, and this issue is going to be relevant in that context. However, there are other points to consider:

1.  I don't expect that complete end-to-end simulations are going to be ran often; in fact, I expect that 99% of the users in LiteBIRD will use our framework just to run specific kinds of simulation. I remember running end-to-end simulations for Planck/LFI, and it was a nightmare to understand the results… too many effects, too complex outputs, it was always very difficult to pinpoint the cause of problems in the final data products, and very often I turned down to re-run simulations with only a few systematic effects turned on. This led to a huge amount of wasted CPU time. I think we should carefully plan how often we're going to run end-to-end simulations in LiteBIRD, and to be sure not to run them too often: ideally, only once we have ran separate simulations for each effect and we're reasonably sure that the results are what we expect, we can think of starting an end-to-end run.

2.   The two approaches to data layout have advantages and disadvantages. The one proposed in this PR is perfect for correlated effects (thermal drifts, cosmic rays, …), while the one in the `master` branch is better for beam convolution. I don't think that we're going to find the best data model that fits any situation, so we should aim to find the one that permits to do most of the stuff adequately. The advantage of this PR is that it can be used similarly to the approach followed in the `master` branch (if you use a Nx1 matrix of samples in each observation, you can easily store a full-length TOD in memory), but it enables other use cases.

dpole:
I've created a page about this data model in the new readthedocs website https://litebird-sim.readthedocs.io/en/multi_det_obs/data_layout.html

It is meant to stay there for reference if and when this PR will be merged

dpole:
The main (only?) missing piece in this PR is how quantities different form the TOD are loaded into the observations. I probably need help from @ziotom78 or @paganol . 
I've prototyped something like
```python
for field in fields:
    field_arr = np.array([some_method(field, d) for d in detectors])
    setattr(self, field, field_arr)
```
in the `Observation` constructor, I'm not sure about how compatible this pattern is with the rest of the framework (and `some_method` is of course to be written).

ziotom78:
> The main (only?) missing piece in this PR is how quantities different form the TOD are loaded into the observations. I probably need help from @ziotom78 or @paganol .
> I've prototyped something like
> 
> ```python
> for field in fields:
>     field_arr = np.array([some_method(field, d) for d in detectors])
>     setattr(self, field, field_arr)
> ```
> 
> in the `Observation` constructor, I'm not sure about how compatible this pattern is with the rest of the framework (and `some_method` is of course to be written).

For pointings ([PR#48](https://github.com/litebird/litebird_sim/pull/48)) I add to create this kind of «field», because the PR introduces pointing information in the form of quaternions just initialize `self.bore2ecliptic_quats` to `None` in the constructor `Observation.__init__` and then I set them once `Observation.generate_pointing_information` is called.

The biggest advantage of this approach over `setattr` is clarity: once you create an object, you immediately have the field names defined in your object, so that once you create an instance of `Observation` like

```python
obs = Observation(…)
```

you can discover the name of the fields by writing `obs.` and hitting `TAB` in Jupyter, but you do not allocate memory for stuff you do not need. Moreover, you can search for these fields in the source code of the class, because they are always initialized explicitly (e.g., just search for `self.bore2ecliptic_quats` in the code to find where the quaternions are generated/used).

dpole:
I see. What you are saying actually simplifies quite a bit the role of this PR: the data layout becomes a convention that users should follow and not something enforced by the `Observation` API.

I think that we should wrap this `Observation` pattern/API in a dedicated issue, as it is an important choice that applies to any information that we store in the `Observation`. I presume it is better if you do it, as you probably have a better overall view, but I'm happy to try to do it myself, if you prefer.

litebird_sim/simulations.py
dpole: > > About performance, all the examples that come to my mind run faster if the the arguments are passed with one array
> 
> You're right, I discovered that creating 100,000 detectors using one list of Detectors takes 90 ms, while creating separate NumPy arrays only takes 50 ms. However, I do not expect that users are going to notice if the code spends 0,04 seconds more in running a function that's only called once in the lifetime of a program… (Moreover, LiteBIRD is going to have far less than 100,000 detectors!)

I didn't have in mind the detector creation, but rather the usage of the information stored in the detectors. 

> 
> > I prefer if we pass only the arguments that the function needs instead of a Detector.
> 
> As I already said somewhere else, I do like this idea of yours that the framework should provide functions with the explicit list of parameters. In fact, my proposal idea is compatible with this code of yours:
> 
> ```python
> sim.create_observation(detectors=detectors, ...)
> for obs in sim.observations:
>     obs.tod += generate_noise(obs.fknee, obs.white_noise, ...)
> ```
> 
> but I mistakenly wrote my example above forgetting `sim.` in front of `generate_noise` (I fixed my comment):
> 
> ```python
> # This expects a list of strings…
> sim.create_observation(detectors=detectors, ...)
> # …but this wants a list of "Detector" object
> sim.generate_noise(detectors=detectors, ...) # Add "sim."
> ```
> 
> The idea is that there is a function, `generate_noise`, that takes the arguments `fknee_hz`, `white_noise_ksqrthz`, etc., exactly like in your example, plus a `Simulation.generate_noise` method that wraps `generate_noise` and uses the IMO to retrieve these parameters coherently. In this way, we have the best of the two worlds. (I am following the same approach in PR #48 : `Simulation.generate_pointing_information` is a high-level wrapper around `Observation.generate_pointing_information` that knows how to access the IMO to retrieve the scanning strategy parameters.)

Not sure it is worth discussing, as it all starts from an example meant for something else. I've got the impression that `generate_noise` or any other similar function don't need detectors as argument, as the simulation and the observations in it already know the detectors they are handling
> 
> > In particular, it seems to me that we could use a simple dictionary (or SimpleNamespace) instead of a Detector.
> 
> My initial implementation of `Detector` was a simple named tuple, but I changed it to a class because in this way we can provide implementations for `Detector.__repr__`, `Detector.__str__`, etc. I agree that this is not as awesome as in Julia, where you can define even a `plot` function over your custom types (see [here](https://ziotom78.github.io/Harlequin.jl/stable/genpointings/) for an example about scanning strategies, and [here](https://lspestrip.github.io/Stripeline.jl/latest/instrumentdb/#Introduction-1) for bandpasses), but this can still be helpful when debugging codes or doing exploratory work..

What about subclassing a dictionary (or SimpleNamespace) and overriding repr and str?
I think that it is important for the users to minimize the number of classes we introduce. I am sure that they will find themselves more comfortable when working with data structures and types they are already familiar with.

I generally agree with the pros you list. About a couple of them, we'll need anyway (bug-free) tools to load and stack the information of the detectors into arrays of the right length and order (that's what this PR is about).

ziotom78:
> I see. What you are saying actually simplifies quite a bit the role of this PR: the data layout becomes a convention that users should follow and not something enforced by the Observation API.

Yes, that's right, I plan to use the same approach for pointings in PR #48 once this one is merged.

> I think that we should wrap this Observation pattern/API in a dedicated issue, as it is an important choice that applies to any information that we store in the Observation. I presume it is better if you do it, as you probably have a better overall view, but I'm happy to try to do it myself, if you prefer.

Well, the most important piece of information in the `Observation` class is the TOD; any other timestream (quaternions, pixel indices, etc.) should follow the TOD. I believe it's ok if we continue the discussion in this PR: it has received so much advertisement in the last weeks that many people in the ST are probably used to check this page.

litebird_sim/simulations.py
ziotom78: > What about subclassing a dictionary (or SimpleNamespace) and overriding repr and str?
I think that it is important for the users to minimize the number of classes we introduce. I am sure that they will find themselves more comfortable when working with data structures and types they are already familiar with.

I am more convinced by the familiarity with data types than by the purpose of minimizing the number of classes we define. (This probably comes from my C++ background, where you're encouraged to create as many classes as the concepts you're dealing with; and Python classes are far easier to declare and use than C++ classes!) However, I bet that everybody going to use our framework is already familiar with the concept of `class`, so `Detector` should cause no harm! :-)

I would avoid dictionaries like the plague, as they can lead to unreadable code (see `detector["noise_{}".format(measurement_unit)]`). The `SimpleNamespace` type is more appealing, but it has a few drawbacks (a few of them shared with a `dict` type):

1. I said above that using many classes is the way C++ programmers write codes. However, this is a deceiving comment, as the way recent versions of Python have evolved seems to acknowledge the importance of quickly creating new classes. In fact, Python 3.7 has introduced [data classes](https://docs.python.org/3/library/dataclasses.html), which seems to have been warmly welcomed by the Python community (see several [blog](https://realpython.com/python-data-classes/) [posts](https://towardsdatascience.com/data-classes-in-python-8d1a09c1294b)) and [HN](https://news.ycombinator.com/item?id=16868296) [comments](https://news.ycombinator.com/item?id=17074783). (I expect that, once we drop support for Python 3.6 at the end of 2021, we should redefine types like `ScanningStrategy` and `Detector` as data classes.)

2. If we decide to stick with «familiar» data types, I fear that `SimpleNamespace` is not familiar at all to the average user of our framework. In fact, it is *never* used in the [TOAST](https://github.com/hpc4cmb/toast), [s4cmb](https://github.com/JulienPeloton/s4cmb), [healpy](https://github.com/healpy/healpy/tree/master/healpy), [PySM](https://github.com/bthorne93/PySM_public) repositories, which are probably quite representative of the kind of code a CMB scientist reads regularly. (The only example I found was *one* line of code in your own [FGBuster](https://github.com/fgbuster/fgbuster/blob/master/fgbuster/observation_helpers.py#L234).) If you google `python simplenamespace`, you get 31,000 results, but if you google `python "data class"`, you get 219,000,000 results!

3. IMO, it provides too much flexibility. The list of fields in a `detector` variable is supposed to be known in advance, as they are the fields our framework is reading from the IMO and is publishing to the user. A `SimpleNamespace` or a `dict` is meant to be used when you want to add/remove keys during runtime; this is not the case for `Detector`.

   In fact, the very [Python documentation](https://docs.python.org/3/library/types.html#types.SimpleNamespace) suggests that you should prefer a `namedtuple` (my first choice for `Detector`, as I said above) over `SimpleNamespace` when you want a structured type, as it is the case for `Detector`.

4. Probably because of its flexibility, a `dict` or a `SimpleNamespace` is hard to document. You can associate a docstring to a `Detector` class, explaining the meaning of each field, the assumptions, examples about how to use its fields, etc. Is the same possible with a `dict` or a `SimpleNamespace`?

5. A `dict` or a `SimpleNamespace` does not allow you to properly specify a type in the declaration of functions and methods: the meaning of the parameter `dets: List[Detector]` is immediate, and if you have doubts about what a `Detector` is, just type `?Detector` in the REPL. How should we declare the parameter `dets` if we use a `dict` or a `SimpleNamespace`? Should we write `dets: List[Dict[str, Any]]` and then add  the list of needed keys to the docstring of that function (of *every* function taking a list of detectors, with the problem of preserving consistency among the comments)?

dpole:
This pull request could be merged at this stage. The missing part is the MPI parallelization. We can merge the PR now and keep working on it.

About the parallelization, I spent quite a bit of time trying to find a python library handling distributed arrays. With my great bitterness, the only one that does simply and effectively what we want (in python) is Dask, which we already discussed and dropped.

The main constraint we have is that we want to be able to pass from one distribution of the TOD to another. Currently the distribution is done by `Simulation` but if we want to rearrange things I believe it should be done in `Observation`. We can pass to its constructor a tuple (n_chunks_detectors, n_chunks_time) and an MPI communicator. The constructor would chunk the TOD matrix in n_chunks_detectors by n_chunks_time blocks, each assigned to a processor of the MPI communicator.

test/mpi/test_mpi.py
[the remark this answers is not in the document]
dpole: This will be fixed once we decide about how to distribute the blocks in MPI enabled runs (see my last comment)

dpole:
I implemented the MPI parallelization as discussed in the previous message:
* the parallelization is done inside the `Observation` object
* an MPI communicator, `n_blocks_det` and `n_blocks_time` are passed at construction and define how the tod matrix is chunked and distributed
* `n_blocks_det` and `n_blocks_time` can be changed at runtime with `set_n_blocks`. It is a quite complicated function (that required a fair amount of suffering), so anybody interested in trying and testing is more than welcome. There are already some test in place.

dpole:
All the key elements of this PR are in place. We have to address the conflicts with master, but I think that the priority is to review the API changes proposed, of the `Observation` class in particular.

https://litebird-sim.readthedocs.io/en/multi_det_obs/observations.html#litebird_sim.observations.Observation

Happy to add clarifications and/or examples either here or in the doc wherever needed.

litebird_sim/observations.py
```diff
@@ -33,18 +41,46 @@ class Observation:
...
+
+        root (int): rank of the process receiving the detector list, if
+            ``detectors`` is a list of dictionaries, otherwise it is ignored.
     """
 
     def __init__(
-        self, detector, start_time, sampling_rate_hz, nsamples, use_mjd=False
+        self, detectors, n_samples,
+        start_time, sampling_rate_hz, use_mjd=False, dtype_tod=np.float32,
+        n_blocks_det=1, n_blocks_time=1, comm=None, root=0
```
dpole, line 66: Add possibility to not allocate tod

litebird_sim/observations.py
```diff
@@ -63,52 +99,372 @@ def __init__(
...
+                info = [info[s:s + n] for s, n, in zip(starts, nums)]
+
+            info = comm_col.scatter(info, root)
+
+        comm_row = comm_grid.Split(comm_grid.rank // self._n_blocks_time)
+        info = comm_row.bcast(info, root_col)
+        assert len(info) == len(self.tod)
+        setattr(self, name, info)
 
     def get_times(self):
```
dpole, line 433: Check that works also if tods are not allocated

litebird_sim/simulations.py
```diff
@@ -364,66 +363,3 @@ def flush(self):
             outf.write(html_full_report)
 
         return html_report_path
-
-    def create_observations(
```
dpole, line 368: Reinsert to allow a second level of time-chunking

litebird_sim/observations.py
```diff
@@ -33,18 +41,46 @@ class Observation:
...
+        comm: either `None` (do not use MPI) or a MPI communicator
+            object, like `mpi4py.MPI.COMM_WORLD`. Its size is required to be at
+            least `n_blocks_det` times `n_blocks_time`
+
+        root (int): rank of the process receiving the detector list, if
+            ``detectors`` is a list of dictionaries, otherwise it is ignored.
     """
 
     def __init__(
-        self, detector, start_time, sampling_rate_hz, nsamples, use_mjd=False
```
ziotom78, line 44: The parameter `use_mjd` is useless in the `master` branch, as the choice to use or not astropy in computing times is done according to the type of the parameter `start_time`. It can be safely removed here.

dpole:
I think that the new layout can be merged and used. There is still some stuff to tidy up (e.g. naming conventions, extract some methods from `Observation` and maybe enhance the doc) but we've decided to keep working on it later.

PR #43, opened 2020-07-06
map based sims implementation 

NicolettaK:
This PR implements the `Mbs` (Map based sims) class. 

The class interfaces with the IMO, reads the channel infos and generates signal (CMB and foregrounds) and white noise maps accordingly, writing them to disk. It makes use of `PySM3` which is therefore added as a dependency in poetry. 

### Examples

This is an example on how to use the class step by step and generate sims:

```python
import litebird_sim as lbs

#read the toml configuration file (see example of file below) and check it
mbs = lbs.Mbs('config.toml')
mbs.check_and_fix_config()

#read instrument from IMO
mbs.read_imo()

#generate and write to disk noise, cmb, foreground maps and coadded signal (cmb+foregrouds) maps
mbs.make_noise_sims()
mbs.make_cmb_sims()
mbs.make_fg_sims()
mbs.coadd_signal_maps()
```

everything can be done just in one step:

```python
import litebird_sim as lbs

mbs = lbs.Mbs('config.toml')
mbs.run_all()

```

example of a configuration file:

```toml
[map_based_sims]

    #general configuration
    [map_based_sims.general]
    parallel = false
    
    #intrument
    [map_based_sims.instrument]
    imo_directory = '/Users/niki/Library/LiteBIRD/litebird_imo/IMOv0.10/'
    imo_version = 'v0.10'

    #output maps parameters
    [map_based_sims.maps]
    nside = 16
    gaussian_smooth = false
    save_coadd = true

    #noise configuration
    [map_based_sims.noise]
    nmc_noise = 2
    seed_noise = 43141
    hm_split = false

    #cmb configuration
    [map_based_sims.cmb]
    cmb_ps_file = false
    cmb_r = 0
    nmc_cmb = 2
    seed_cmb = 4984

    #foregrund configuration
    [map_based_sims.fg]
    dust = 'pysm_dust_0.cfg'
    synch = 'pysm_synch_0.cfg'
    freefree =  'pysm_freefree_1.cfg'
    ame =  'pysm_ame_1.cfg'

    #output options
    [map_based_sims.output]
    out_directory = './'
    file_string = 'example'
```
-----------------------
It would be great if you could have a look at it and provide comments.

Several things still need to be implemented:

- [ ] fix the use of mpi4py (what is in the code right now is old, at the moment parallelization in not implemented)
- [ ] implement the report writing
- [ ] documentation


ziotom78:
Hi @NicolettaK , thanks a lot for this! This is the very first module to be integrated in the framework, so we're going to learn a lot from this PR.

A couple of things:

1.   My idea is that we should save the configuration for the IMO (i.e., `imo_directory` / `imo_server` / `imo_username`, etc.) in a separate configuration file, to be kept in the user's home directory, instead of specifying it in every parameter file we're passing (as you're doing . However, it seems that this idea is controversial, because of some comments by an anonymous user in the «Requirement document for the LiteBIRD Simulation Framework». Perhaps it's better to open a dedicated issue on this.

2.   I see that one thing you want to implement is the generation of reports: good! To do this, you should initialize a `Simulation` object in your example, and use it to write reports:

     ```python
     from litebird_sim import lbs
     sim = lbs.Simulation(base_path="/my/simulation/results")
     sim.append_to_report(…)
     ```

     This makes the section `map_based_sims.output` in the TOML file redundant, as you're using the `base_path` parameter when constructing the `lbs.Simulation` object. Moreover, you won't have to initialize the IMO, as it is already available in `sim.imo`. (Well, no, not at the moment, you would need to pass `imo=…` to `Simulation`'s constructor; but if we implement point 1. above, this could be done automatically.)

3.   I think that the best option would be to specify the TOML file in the `lbs.Simulation`'s constructor: in this way the TOML could be copied automatically in the output directory and included in the final report. It could be accessed through `sim.parameters`, or `sim.configuration`. Perhaps it's better to open a dedicated issue for this as well?

NicolettaK:
Thanks @ziotom78!

In general I strongly believe it would be better to have just one TOML file with the configuration for all the modules, including the IMO one. 

I think I need you help to understand how to fully integrate this with the `Simulation` object. As you said this is the first case of integration of a module in the framework and it will outline the standard procedure.  It is therefore important to do it in what we all think is the right way.

ziotom78:
> In general I strongly believe it would be better to have just one TOML file with the configuration for all the modules, including the IMO one.

In general I agree with you, as keeping everything in one file ensures reproducibility. However, this case is peculiar and having a separate file for the IMO configuration has several benefits:

1. You might want to specify a remote IMO instead of a local file, and in that case you must put your username and password in the TOML file. Consider that the plan is to implement a feature to copy the TOML file in the output directory and append its contents to the report: this means that your password would be always copied in the report in clear text.
2. It makes the TOML file easier to port on other architectures, as it does not refer to some directory in your computer. (This is handy in several situations, e.g., you test a simulation on your laptop and then move the TOML file to a HPC cluster for a bigger run.) However, TOML files will be perfectly portable only once we have a Sky Model implemented in the same way as an Instrument Model, because at the moment we need to refer to foreground models using lines like `dust = 'pysm_dust_0.cfg'`, which refer to local files 😟.
3. It eases the configuration, as the user only needs to configure the IMO once.

NicolettaK:
Ok, we can discuss better how to structure the parameter file(s) when we will have more than one module to configure.

> However, TOML files will be perfectly portable only once we have a Sky Model implemented in the same way as an Instrument Model, because at the moment we need to refer to foreground models using lines like dust = 'pysm_dust_0.cfg', which refer to local files 😟.

sky models are actually here: https://github.com/litebird/litebird_sim/tree/map_based_sim/litebird_sim/mbs/fg_models
they shouldn't be a problem!

mreineck:
> You might want to specify a remote IMO instead of a local file, and in that case you must put your username and password in the TOML file.

Sorry for not spotting this earlier ... but please let's not design anything which requires putting unencrypted passwords on disk. Computing centers take a very dim view on this practice, especially since many of them had to shut down for quite some time recently because users did exactly this.

ziotom78:
> sky models are actually here: https://github.com/litebird/litebird_sim/tree/map_based_sim/litebird_sim/mbs/fg_models
> they shouldn't be a problem!

No, I was referring to the **software stack** for the Sky Model: a software layer that can translate a string like `/releases/v0.1/diffuse_foregrounds/dust_model` into a true file. A software like `instrumentdb` is perfect for this kind of stuff (as it does not differentiate between "instrument" and "sky" objects), but nobody has agreed that this is required, yet.

ziotom78:
> Sorry for not spotting this earlier ... but please let's not design anything which requires putting unencrypted passwords on disk. Computing centers take a very dim view on this practice, especially since many of them had to shut down for quite some time recently because users did exactly this.

People are not going to use remote servers for the IMO on computing centers, as the plan is to have local copies there to ensure good performance. Remote servers will be used mainly by people developing codes or running short pipelines on their laptops. There are ways to mitigate the risks, from the easiest to implement (e.g., enforcing 0600 permission to the configuration file) to the hardest (using private/public RSA/DSA keys), but it's better to talk about this in a new dedicated issue.

PR #46, opened 2020-07-18
Implement TOML file loading in Simulation objects

ziotom78:
This PR adds a parameter, `parameter_file`, to the constructor of the class `Simulation`. This must be a path to a TOML file, which is meant to contain the parameters to be used in the simulation.


dpole:
Hi Maurizio (and everyone), I'd like to propose the following conventions about the parameter file. I think they are compatible with your proposal. I number them to ease referencing, not because they are ranked by importance. Let me know what you think!

1. Every pipeline is a (nested) list of steps. The parameter file should reflect this in a nested hierarchy of tables
2. If an analysis step expects a parameter that is not found in the corresponding table, the hierarchy of parameters is walked backwards until the parameter is found. Therefore, if many analysis steps require the same parameter, it can be defined once for all in the lowest branch of the hierarchy shared by the steps
3. If a key of a table is an imo version, it will be searched for missing arguments before jumping to the hierarchy above. Note that this means that

    * different analysis steps can use different IMOs
    * if you define one IMO in the top table you can avoid specifying frequencies, sensitivities, etc. for any step of the pipeline

ziotom78:
Hi @dpole , thanks for these ideas. Let me try to produce an example:

```toml
[simulation]
nside = 128
imo_version = "v1.0"
detectors = ["lft_40ghz", "lft_50ghz"]

[simulation.scan_sky]
scanning_strategy = "/releases/{imo_version}/Satellite/scanning_parameters/"

[simulation.add_noise]
# There should be a "detectors = […]" line here, but it's missing
detector_noise = "/releases/{imo_version}/Satellite/{detector}/info/"

[simulation.map_making]
# There should be "nside = NNN" somewhere in this section, but it's missing
```

The example should simulate a scanning strategy for the two detectors `lft_40ghz` and `lft_50ghz`, using the object `/releases/v1.0/Satellite/scanning_parameters/`, add some noise and produce a map using `nside=128`.

Considering that the `scan_sky` and `add_noise` modules could provide the two long IMO strings above as defaults, the example above might be written more concisely in this way:

```toml
[simulation]
nside = 128
imo_version = "v1.0"
detectors = ["lft_40ghz", "lft_50ghz"]

[simulation.scan_sky]

[simulation.add_noise]

[simulation.map_making]
```

Does this translates faithfully what you were proposing in your comment?

dpole:
Yes, very nice example. Thanks.
What about skipping the `simulation` and letting the steps as 0th level keys?

ziotom78:
Ok, thanks for the clarification. I like this approach, it makes the configuration file much easier to read. However, there is a point that I believe needs to be discussed a bit more.

I begin to dislike the specification of IMO objects through plain strings like `/releases/v1.0/Satellite/…`, because it does not permit to distinguish between local file paths and proper IMO objects. Suppose that some code needs to take a Grasp beam file as input. When the code reading the TOML parameter file encounters the string `/releases/v1.0/Satellite/beams/lft_30GHz_foo`, how is it supposed to know if it points to a local file in the user's hard disk or a real IMO object? This issue is related to this PR, because the kind of string substitution you are suggesting in point 3 of your proposal should work only for IMO references, and not for local file paths.

Perhaps the cleanliest way to operate here is to require IMO paths to begin with an [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) scheme specification, like `imo://`. Thus,

-   `imo:/releases/v1.0/Satellite/beams/foo` would be a reference to an IMO object;
-   `file:///releases/v1.0/Satellite/beams/foo` would refer to file `foo` in the local directory `/releases/v1.0/Satellite/beams/foo`;
-   `/releases/v1.0/Satellite/beams/foo` would be the same as having `file://` prepended at the beginning.

Point 3 in your proposal (automatic substitution of keys in IMO URIs) would thus be easily implemented, as it would work only if strings in the TOML file begin with `imo:/`.

dpole:
I can easily miss some important technical detail about the IMO. My naive idea about point 3 didn't involve a string substitution. Whenever the parser finds the key `imo` (I think we can drop the `_version`) it creates an IMO object that points to the specified version. If you want a parameter to be loaded from imo, you don't specify anything. The pipeline stage should know how to properly query the imo. If the query need other parameters, like some detector identified, it will look for it recursively, just like if it was a parameter.

It is getting cumbersome, I agree. I just would like that the used doesn't have to specify imo paths, if possible.

ziotom78:
> Whenever the parser finds the key imo (I think we can drop the _version) it creates an IMO object that points to the specified version.

Mmm, in its present stage, the IMO object cannot "point" to some specific version. They way "releases" are implemented is just as a bunch of UUIDs, but perhaps I did not understand what you're suggesting.

> It is getting cumbersome, I agree. I just would like that the used doesn't have to specify imo paths, if possible.

I fear that providing too much abstraction might lead to a solution that is more complicated than the original option. Keep in mind that dealing with IMO paths will be considerably easier once a proper web interface is available. When this will happen, it might actually be *handier* to have IMO paths in a configuration file than not to have them.

Consider a real-life case. You stumble upon a configuration file that was in your hard drive since some time, and you want to recall what it did:

```toml
nside = 128
imo_version = "v1.0"
detectors = ["lft_40ghz", "lft_50ghz"]

[scan_sky]
scanning_strategy = "/releases/{imo_version}/Satellite/scanning_parameters/"

[add_noise]
# There should be a "detectors = […]" line here, but it's missing
detector_noise = "/releases/{imo_version}/Satellite/{detector}/info/"

[map_making]
# There should be "nside = NNN" somewhere in this section, but it's missing
```

You could easily open the IMO website and type `/releases/v1.0/Satellite/scanning_parameters/` in the search bar (once the webpage and this search bar will be implemented!): a page showing the details of the object will appear, possibly including even a plot.

OTOH, if everything is implied in the configuration file, it means that you won't be able to know what are the actual parameters of the simulation once you fully run it and then open the report containing the list of accessed IMO objects (potentially navigating through several thousands of them!), looking for some path that looks like `scanning_parameters`, `scanning_strategy`, or something like (because you have no clue what's the actual path of the object).

I'm not advocating *against* letting the code sort out which IMO paths to use, if they're not specified. Sometimes it's very handy to quickly put together a configuration file, like the one I wrote above:

```toml
nside = 128
imo_version = "v1.0"
detectors = ["lft_40ghz", "lft_50ghz"]

[scan_sky]

[add_noise]

[map_making]
```

But there are different tastes around, and some people (including myself) prefer to have everything explicit when they produce their own configuration files than to rely on defaults. Defaults can change! For instance, we'll likely use top-hat bandpasses parametrized by Δν and ν0 until we have realistic simulations/measurements; but once we have them, we'll forget about Δν and ν0 and will only use true bandpasses. As top-hat bandpasses and realistic ones will live in different branches of the IMO tree, they will be accessed using different paths. If IMO paths can only be implicit, there will be a potential source of confusion here: if you know some module uses bandpasses, which one is it going to use? Will it depend on the release date of `litebird_sim`? (In the sense that older versions will use the path to top-hat bandpasses, and newer will point to the true bandshapes.)

Now let's consider another example, from a different angle this time.

Imagine that we have a class that generates noise, following the example above with the `add_noise` section:

```python
class AddNoiseModule:
    def __init__(self, sim):
        detector_noise_imo = sim.parameters["add_noise"].get(
            "detector_info_imo",
            "/releases/{imo}/Satellite/{instrument}/{detector}/info/",
        )
        # Here the code should expand {imo}, {instrument}, and {detector}
        self.detector_noise = sim.imo.query(detector_noise_imo)

    def run(self):
        ...
```

With this implementation, the user can specify arbitrary paths to IMO objects (including explicit UUIDs!), and yet the code is able to use sensible defaults, which means that the following section in the TOML file is still valid:

```toml
[add_noise]
# Empty
```
I believe that this implementation should be the way to go: it's easy to use, but it allows people enough flexibility to tailor their configuration files as they want.

dpole:
Thanks for the details, I see your point --which fits very well in your general prioritization of reproducibility and traceability. I am wondering whether we can find a way to also keep simple things simple. 

For example, suppose that one just wants to get the reference frequencies and white noise levels in the maps. Do you think we can find a way to avoid requiring the user to look up non-trivial paths, iterate over channels, etc? This is really more an IMO question than a parameter file question. To be even more explicit. Suppose I want the instrument version v28.20. I'd like to be able to do something as simple as `v28 = get_me_this_config('v28.20'); depth = v28.depth; freqs = v28.freqs`

dpole:
Another comment, which we discussed offline but I'd like to make it also here for the general records.

About your AddNoiseModule example, I would prefer if constructors of pipeline stages didn't take a Simulation as argument. From one hand, I agree that is much more convenient to pass around only one object that knows everything, from the other I'd prefer if we passed only the specific objects needed by the module. For example, I would like to develop, use interactively, write tests for a module without defining a Simulation objects and knowing how it works.

[We can discuss it on Thursday or move this to a separate issue, if you prefer]

ziotom78:
> Do you think we can find a way to avoid requiring the user to look up non-trivial paths, iterate over channels, etc?

Yes, I thought about this issue. @paganol devised a nice solution for its flat-file IMO, where he included a "quantity" including the full list of detectors at a specified frequency. The IMO could adopt a similar approach, and the user might decide whether to specify the name/IMO path to a *single* detector, or to one of these list-like quantities.

ziotom78:
As we discussed during the last telecon of the Simulation Team, the PR as it is now can be merged. The features discussed in this thread are going to be integrated later using a separate PR.

PR #47, opened 2020-07-18
Automatically configure the "logging" library

ziotom78:
This PR adds a call to `logging.basicConfig` when constructing a `Simulation` object. The documentation has been updated accordingly.


ziotom78:
I made a mistake, the tip of this PR is not on `master` but on the `conffiles` one (see PR [46](https://github.com/litebird/litebird_sim/pull/46)). I'll leave it as it is; this means that this will be merged once PR46 is closed.

PR #48, opened 2020-08-01
Implement a pointing generator

ziotom78:
This PR implements a pointing-generation code, i.e., a set of routines that take a scanning strategy definition as input and produce a set of direction in the sky and polarization angles as output. Once the PR is finished, I expect it to be used in the following way:

```python
import litebird_sim as lbs

sim = lbs.Simulation(…)
sim.create_observations(…)
sim.generate_pointing_information(imo_object="/releases/v1.0/Satellite/scanning_information/")

# Now pointing information are available in each "Observation" object
```

The method `generate_pointing_information` works as follows:

1. It creates a set of quaternions representing the transformation from the boresight reference frame to Ecliptic coordinates; this is kept in a matrix whose shape is `N x 4` and using the scalar-last convention, which is compatible with the data layout used by `ducc` (see [github.com/litebird/ducc/pull/1](https://github.com/litebird/ducc/pull/1)).
2. Depending whether the flag `use_mjd` has been set or not in the call to `sim.create_observations`, the transformation to Ecliptic coordinates uses a [barycentric true ecliptic](https://docs.astropy.org/en/stable/api/astropy.coordinates.BarycentricTrueEcliptic.html) coordinate system and the true Sun-Earth direction (using [astropy.coordinates](https://docs.astropy.org/en/stable/coordinates/solarsystem.html)) or a quick-and-simple geometrical representation of the Ecliptic plane.
3. The set of quaternions is calculated at a smaller sampling rate than the scientific samples. At the moment they are sampled once every 60 s, which means that 4 years of mission require ~64 MB for all the quaternions, but this parameter is tunable.
4. When the user needs pointings, they are calculated using the method `PointingProvider.get_rotated_quaternions` implemented in `ducc` (see [this example](https://github.com/litebird/ducc/blob/ducc0/python/test/test_pointing.py)).
5. It is possible to avoid using a `Simulation` object and generating the pointings for just one observation using `Observation.generate_pointing_information`. This can be handy for debugging and is already used in the unit tests.

Things to do:

- [x] Implement basic quaternion operations
- [x] Implement a `ScanningStrategy` object
- [x] Compose quaternion operations to convert from the boresight reference frame to the Sun-Earth axis frame
- [x] Apply the appropriate conversion from the Sun-Earth axis frame to the Ecliptic frame, using ephemerides whenever `use_mjd=True`
- [x] Write tests to check that quaternions are generated correctly
- [x] Write the code that retrieves the scanning strategy from the IMO
- [x] Implement `Simulation.generate_pointing_information` so that it iterates over all the observations and calls `Observation.generate_pointing_information` for each of them
- [x] Use `ducc` to generate pointing information and polarization angles at the same sampling rate as the data
- [x] Write tests to check that the pointing information and polarization angles are correct
- [x] Write docstrings
- [x] Write the documentation
- [x] Provide a way to use different boresight definitions (LFT/MFT/HFT)
- ~~Find a way to fix the warnings produced by ERFA about «dubious years» (see [github.com/astropy/astropy/issues/5809](https://github.com/astropy/astropy/issues/5809) for an explanation of the cause)~~


ziotom78:
@mreineck , I am having problems compiling `ducc` on Mac OS X, see here: https://travis-ci.com/github/litebird/litebird_sim/jobs/367296483. I discovered that it does not compile using Ubuntu 18.04 LTS because the GCC is too old, and I suspect that the issue we're experiencing here with Mac OS X has the same origin.

If you tried to set up a Travis CI job for the library, can you share your scripts so that I can integrate your solution here? At the moment the PR does not use `ducc`, but it will do it soon.

mreineck:
Are you actually using `gcc/g++` as the compilers here? The Travis log only mentions `clang`, and also I'd never expect `gcc` to say something like ` XYZ is unavailable: introduced in macOS 10.14`. I don't understand the statement that the OS somehow determines which language features are supported by a compiler; I can only assume that the system's default compiler is not new enough to support C++17. But there should be no problem installing more recent compilers.

Any `g++` and `clang` with C++17 support should do; ccording to Wikipedia this is g++ 7 or later and clang 5 or later (but the MacOS system clang seems to follow a different versioning scheme than the official one ...)

I have set up CI with Gitlab, will attach the relevant files in a minute!

mreineck:
No need to upload anything ... the files are `Dockerfile` and `.gitlab-ci.yml` in the `ducc` repo.

ziotom78:
> Are you actually using gcc/g++ as the compilers here?

I have just asked TravisCI to install XCode and use whatever is available; I think that this makes the VM defaulting to `clang`. I am going to bump the requirement on XCode from 11.0 to 12.0, let's hope that this fixes the build.

> I don't understand the statement that the OS somehow determines which language features are supported by a compiler; I can only assume that the system's default compiler is not new enough to support C++17

It sounds troublesome at best… I found nothing else than a Reddit topic about this: https://www.reddit.com/r/cpp_questions/comments/9la2e7/technical_reasons_for_apple_restricting_any_and/

ziotom78:
Ok, this build failed too:

https://travis-ci.com/github/litebird/litebird_sim/jobs/367303706

It seems you're right: even if the SDK is the most recent one, it fails to build because the OS does not support some functionalities required by the C++ library, as this error message suggest:

    /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/malloc/_malloc.h:50:10: note: 'aligned_alloc' has been marked as being introduced in macOS 10.15 here, but the deployment target is macOS 10.9.0

(So Mac OS X 10.9 doesn't provide functions to allocate a block of aligned memory… really?!?)

However, this is in conflict with the TravisCI documentation (https://docs.travis-ci.com/user/reference/osx#macos-version), which states that specifying `xcode12` in the Travis script will use Mac OS X 10.15.

I think we need some help from people using the Mac. @dpole , might you please try to download the `pointings` branch and check that you can run the tests?

mreineck:
> (So Mac OS X 10.9 doesn't provide functions to allocate a block of aligned memory… really?!?)

I'm pretty sure it provides the function, but htis message is about the `aligned_alloc` function, which was added in C++17 and is part of the C++ standard library.
Is it possible that on MacOS, all C++ binaries are linked to the _system provided_ C++ standard library, and not the one that comes with the compiler? If so, this message makes at least remote sense, because then it wouldn't matter how new your `clang` is, because in order to use `aligned_alloc` you also need a recent standard library. And if the custom is to link to the outdated, system-provided one, this will fail.

In any case, I cannot imagine that this kind of trickery is going on when `g++` is used.

mreineck:
BTW, I'm sorry that this code is proving such a hassle to compile... after your success on Windows, I definitely didn't expect this. I'm also building the library regularly with `clang` under Linux, and that also works perfectly well.
(OK, to be honest, I had to switch off CI tests with `clang` recently, but that's because Debian testing currently has their dependencies wrong. That should be fixed soon, however.)

ziotom78:
> Is it possible that on MacOS, all C++ binaries are linked to the system provided C++ standard library, and not the one that comes with the compiler?

Yeah, I believe so, this would explain everything. What puzzles me is the fact that Travis should be using Mac OS X 10.15, not Mac OS X 10.9.

> BTW, I'm sorry that this code is proving such a hassle to compile...

Don't worry, this is the very first C++ library we're integrating in the code, so it makes sense that we're experiencing many problems. Once we're able to understand how to make it work, I expect that other C++ libraries will be far simpler to use.

mreineck:
> What puzzles me is the fact that Travis should be using Mac OS X 10.15, not Mac OS X 10.9.

I wonder if this is related to the fact that my `setup.py` required a minimum version of 10.9...
This is now updated to 10.15 ... with a bit of luck it could work now.

mreineck:
Yes! I think that did the trick :)

ziotom78:
> Yes! I think that did the trick :)

You're my hero, thanks a lot!

mreineck:
I still don't feel particularly good about requiring such a recent OS version... I mean, recent compiler versions are easier to install than newer OSs. What do you think, should I try to avoid the `aligned_alloc` call or can we reasonably expect that our users will be on MacOS 10.15?

ziotom78:
> What do you think, should I try to avoid the aligned_alloc call or can we reasonably expect that our users will be on MacOS 10.15?

Mmm, this depends on how often Apple users upgrade their OS. I guess they do it promptly, but it's probably better to ask to other members of the ST that are Apple users if this is true (I know of @NicolettaK , @paganol , @dpole , but there are surely many others).

This problem is going to be less and less relevant as time passes, considering that Apple has just announced the successor to 10.15, and that Apple is famous for quickly dropping support for older OSes. ([Wikipedia](https://en.wikipedia.org/wiki/MacOS_version_history#Releases) extrapolates old data to guess that 10.14 will be supported till September 2021, i.e., one year from now.) I would go for dropping support for Mac OS X 10.14 and earlier versions, but let's hear for other opinions.

NicolettaK:
Ehm, I've just checked and I still have Mac OS X 10.14...

ziotom78:
> Ehm, I've just checked and I still have Mac OS X 10.14...

Argh, then perhaps you could use something like

```c++
#ifdef __APPLE__
…
#else
…
#endif
```
to use `aligned_alloc` only on non-Apple platforms, at least until 10.14 goes out of support. (Don't know if more sophisticated tests that check for 10.15 and newer are worth the effort.)

mreineck:
I'm trying exactly this right now. The problem is that C++17 not only provides `aligned_alloc`, it also guarantees that data types which require special alignment (like SIMD types) will autimatically get them, e.g. if I have a `std::vector<simd_type>`, that the allocated memory for the `vector` will be sufficiently aligned. No such guarantee was made before C++17.

So if Apple could not be bothered to introduce the `aligned_alloc` function before MAcOS 10.15, I'm not too confident that they will have implemented this subtle language change...

mreineck:
OK, I'm falling back to some emulation of `aligned_alloc` when `__APPLE__` is defined, and I reduced the version requirement to 10.14. So far it seems to work.

mreineck:
I have added Travis CI to `litebird/ducc`, testing MacOS 10.14 and 10.15. Both are passing.

dpole:
Hi everyone, I'm catching up late with this, so I'm not sure whether this is relevant anymore.
I tried running tests and got no problems. I have OSx 10.15.5

dpole:
> Is it possible that on MacOS, all C++ binaries are linked to the _system provided_ C++ standard library, and not the one that comes with the compiler? If so, this message makes at least remote sense, because then it wouldn't matter how new your `clang` is, because in order to use `aligned_alloc` you also need a recent standard library. And if the custom is to link to the outdated, system-provided one, this will fail.

Is this option still on the table? If so, I 'm wondering whether this could be related to the fact that we install poetry with

```
curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python
```

instead of 

```
curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | python3
```

I remember I saw some posts advocating that the former can create conflicts with the system-default python.

ziotom78:
> I 'm wondering whether this could be related to the fact that we install poetry with…

Interesting comment. @mreineck , I did not understand if you got errors in your repository too (which does not uses `poetry`). If everything went on smoothly on your side, perhaps this is something we could fix in the scripts and our README.

mreineck:
I did not get any errors in my repository, but I only enabled Travis after I had made all the adjustments mentioned before (i.e. not using `aligned_alloc` under MacOS, and requiring MacOS 10.14 for support of `std::any`).

I'm using `pip3` in `.travis.yml` to install things. If I should try anything else, let me know!

litebird_sim/scanning.py
```diff
@@ -0,0 +1,399 @@
+# -*- encoding: utf-8 -*-
```
dpole, line 1: In my experience it always turned out to be very useful to allow arrays as inputs and follow the numpy broadcast rules. What do you think about it? I seems to me that the current API does not allow for that. I provide some examples below

litebird_sim/scanning.py
```diff
@@ -0,0 +1,399 @@
...
+    return (0.0, np.sin(theta / 2), 0.0, np.cos(theta / 2))
+
+
+@njit
+def qrotation_z(theta):
+    return (0.0, 0.0, np.sin(theta / 2), np.cos(theta / 2))
+
+
+@njit
+def quat_right_multiply(result, other_v1, other_v2, other_v3, other_w):
```
dpole, line 29: Allow `result` and `other` (I use the entire a quaternion instead of the individual components) to be arbitrary stacks of quaternions the following code should do the job
```python
    v1 = (
        result[..., 3] * other[..., 0]
        + result[..., 0] * other[..., 3]
        + result[..., 1] * other[..., 2]
        - result[..., 2] * other[..., 1]
    )

.......

    result[..., 0] = v1
    result[..., 1] = v2
    result[..., 2] = v3
    result[..., 3] = w
```

dpole: In this example `result` can have any shape `(a, b, 4)` and `other` can have any shape broadcastable to the one of `result`

mreineck:
Maybe not entirely orthogonal to @dpole's coments: `scanning.py` may have quite some overlap with the recently added functionality in `scipy.spatial.transform.Rotation` (https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.transform.Rotation.html). Using this functionality where possible should already cover the automatic broadcasting mechanism, and it might also be faster than a pure Python implementation.
Having said that, I admit that I didn't check how large the overlap actually is...

ziotom78:
Hi @dpole and @mreineck , thanks for the comments! A few of the functions in this PR (which is still WIP) are not meant to be used by the end user (like `quat_right_multiply`), and a few of them are useless and will be removed (`rotate_vector`, `all_rotate_vector`). This kind of low-level functions is kept as simple as possible to help Numba in function inlining, although I have not checked yet how efficient is the code it generates. (I hope it is not much inefficient, I avoided using broadcasting to increase data locality in computations.)

Regarding `scipy.rotation`, I wasn't aware of it. Consider however that the scanning strategy generator only uses `rotate_x_vector`, `rotate_x_vector`, and `rotate_x_vector`, which are optimized for the special cases where you rotate around the x, y, and z axis. It doesn't seem that `scipy.rotation` provides the same.

About broadcasting, it would be completely useless for the current implementation, which does not use broadcasting anywhere. But if you believe it could be useful for the end user, I can add the functionality to `quat_right_multiply` (@dpole, when you write `...` in your example code, are you using an actual feature of Numba? I was not aware of this syntax.)

dpole:
> Hi @dpole and @mreineck , thanks for the comments! A few of the functions in this PR (which is still WIP) are not meant to be used by the end user (like `quat_right_multiply`), and a few of them are useless and will be removed (`rotate_vector`, `all_rotate_vector`). This kind of low-level functions is kept as simple as possible to help Numba in function inlining, although I have not checked yet how efficient is the code it generates. (I hope it is not much inefficient, I avoided using broadcasting to increase data locality in computations.)

Yes, I am not a numba expert and can not tell what level of numpy numba is able to digest. 
About locality, I don't think that braodcasting should negatively affect it. I actually think it improves it. If locality is a concern, I'd start by requiring quaternions to be in a length-4 array (I would do that in any case)

> 
> Regarding `scipy.rotation`, I wasn't aware of it. Consider however that the scanning strategy generator only uses `rotate_x_vector`, `rotate_x_vector`, and `rotate_x_vector`, which are optimized for the special cases where you rotate around the x, y, and z axis. It doesn't seem that `scipy.rotation` provides the same.

I wasn't aware either until this morning, when I started googling because I couldn't believe that there wasn't any standard library for quaternion operations. I should say that I'm not sure it is mature enough to be considered standard but, skimming through the source, it seems to me that all the functionalities we need are there. Rotator is essentially a quaternion container with a bunch of methods for compositions, representation, rotation application etc.

> 
> About broadcasting, it would be completely useless for the current implementation, which does not use broadcasting anywhere. But if you believe it could be useful for the end user, I can add the functionality to `quat_right_multiply` (@dpole, when you write `...` in your example code, are you using an actual feature of Numba? I was not aware of this syntax.)

From a quick look I believe that it could be useful already in the current implementation. I presume that broadcasting would percolate from `quat_*_multiply` all the way to `all_boresight_to_ecliptic` and make it redundant, as broadcasting would automatically take care of the for loop.

`...` in the indexing is a standard numpy indexing thing, not numba's. It allows indexing the last dimension of an ndarray
The only advantage that my proposal would have over the `scipy.spatial.transform.rotation._compose_quat` is that it would allow any broadcasting, not just 2-D broadcasting (Rotation allows only for Nx4 quaternions)

ziotom78:
> About locality, I don't think that braodcasting should negatively affect it. I actually think it improves it.

Mmm, I'm doing a few tests using broadcasting and the `scipy` rotation matrices, but my Numba code seems to be ~2 times faster in calling `all_compute_pointing_and_polangle` (source code in [this gist](https://gist.github.com/ziotom78/84e1d42c0c777896433dfb8639eda2ce)):

```
In [1]: %run scipy_rotations_test.py
Consistency test: True

In [2]: %timeit test_lbs(quat_matrix)
1.55 ms ± 12.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [3]: %timeit test_scipy(quat_matrix)
3.48 ms ± 18.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

Introducing broadcasting without causing serious performance regressions does not seem trivial; at least, not for me. I don't have time to investigate this, and the Numba code I wrote so far works well (all the results look mathematically correct), so for the moment I'll leave everything as it is. We can always introduce broadcasting later, once we find the culprit.

dpole:
From what I could understand you are moving some data from the observations to other objects.

I believe that it would be good the keep the observation as the only class responsible for the distributed of time-domain or detector-domain data (actually I would use this as the definition of what the Observation class should do). The idea is to keep the complexity of chunking and distributing this information in one place (the Observation class) and have other classes (possibly functions) that take data from the observations and operate over it.

What do you think?
[I skimmed through your latest changes quickly and may have missed something, sorry about that]

dpole:
About quaternions routines, I'm fine with that. The scipy implementation doesn't look like the most efficient thing one can do and your test demonstrates it. What I had in mind was to write broadcast-compatible versions of your routines, but I'd keep it as a low priority thing. I'll propose a patch if I have time.

ziotom78:
> I believe that it would be good the keep the observation as the only class responsible for the distributed time-domain or detector-domain data.

I agree. In fact this is still the approach used in the latest commits.

ziotom78:
I believe the PR is now ready to be merged.  The API will need to be revised once #42 is merged, but this can be done with a new PR. Please have a look at it, the documentation is available here:

https://litebird-sim.readthedocs.io/en/pointings/scanning.html

PR #56, opened 2020-09-12
Retrieve simulation parameters from TOML file

ziotom78:
This PR enables to read a few parameters needed to construct a `Simulation` object from a TOML file. As stated in the modified  documentation for this PR, the following code:

```python
sim = Simulation(
    start_time=astropy.time.Time("2020-02-01T10:30:00"),
    duration_s=3600.0,
    name="My simulation",
    description="A long description should be put here")
)
```

is equivalent to this code:

```python
sim = Simulation(parameter_file="foo.toml")
```

provided that file `foo.toml` contains the following `simulation` section:

```toml
[simulation]
start_time = 2020-02-01T10:30:00
duration_s = 3600.0
name = "My simulation"
description = "A long description should be put here"
```


ziotom78:
I have tested this new approach intensively in the last days in real-life scripts, and I find it nice and smooth. Since I don't believe that these changes are controversial, I'm going to merge the PR right now. Feel free to open an issue if you think there is something

PR #58, opened 2020-09-23
Automatically download and configure the IMO

ziotom78:
This PR implements a way to automatically configure an `Imo` object through a configuration file saved in the user's home directory.

The PR adds a standalone script, `litebird_sim/install_imo.py`, which can be run with the syntax

```
python -m litebird_sim.install_imo
```

The script prompts the user interactively to configure the IMO:

install_imo

At the end, the script creates a configuration file in the user's home directory. From now on, creating an `Imo` object with no arguments will use the parameters saved in the configuration file, e.g.:

```python
import litebird_sim as lbs

imo = lbs.Imo()    # Use the configuration file
```

Even better, when you create a `Simulation` object, an `Imo` object will be silently created and made available:

```python
import litebird_sim as lbs

sim = lbs.Simulation()

# You can use `sim.imo`, as it has already been configured
```


ziotom78:
Given that the changes to the existing code are minor, that there are no breaking changes, and that this PR was already discussed during the last telecon, I'm merging this now. Feel free to open issues and create PRs to add missing functionality.

Issue #53, opened 2020-09-08
Do we really want `Detector`?


dpole:
Hi all, I believe that the `Detector` class adds more complications than benefits. Yes, I know that this topic was already discussed in #42 , but let me try again or at least summariz that discussion in a dedicated issue (even I had a hard time looking up the #42 PR)

Benefits are (@ziotom78 please feel free to add/amend in case a missed or misunderstood anything)

1. detectors are an important concept, they deserve a class
2. the class can provide custom `__repr__` and `__str__`
3. the class can enforce the attribute names, which helps the coherence across the framework
4. since the `Detector` stores all the possible piece of information about the detectors, the user always have them at his/her disposal (for example when adding a new stage to an existing pipeline)
5. In spite of 4., at construction time we can specify only the parameters we need: all the rest is defaulted to `None`

Compared to a simple dictionary, I found the following complications
<ol type="a">
  <li>I think that the benefits of 4 are largely outweighed by the fact that every user will be interested in a small subset of quantities, probably different for every user. Still, if any user at any time will need a detector quantity, we will have to add it to the `Detector` attributes. The result is that *we force all the users to carry a lot of quantities even if they only care about a few*.</li>
  <li>I think that *the dictionary is a more natural and transparent data structure*, which the user find more comfortable with. The following points are examples of this general idea</li>
  <li>Sometimes you want to iterate over the quantities: it is more natural if the detector is a dictionary</li>
  <li>Getting all the available detector quantities is much more natural ( det.keys() ) if  det  is a dictionary</li>
  <li>it happens that the name of the quantity you want to access is stored in a variable, I prefer  det[variable] over  getattr(det, variable) </li>
</ol>

dpole:
About the benefits, 1. is very popular idea in the c++ community, but I don't think it has practical advantages so I wouldn't use it as an argument. 
About 2, true, but I wouldn't consider it a driver. 
About 3, it does not sound so different from providing a list of names that people should conform with.

ziotom78:
Thanks for having opened this, @dpole! Before discussing your list, it's better to state a few facts:

-   The `Detector` class is meant to closely match the IMO quantity `(detector) info`: there should be a one-to-one relationship to the metadata in the IMO quantity and the fields in the `Detector` class. (This is actually the case with the current implementation in `master`.)

-   To remark the association, `Detector` should probably be renamed to `DetectorInfo`.

[...]

PR #60, opened 2020-09-23
Use "dataclasses"

ziotom78:
This PR adds the support for
[dataclasses](https://docs.python.org/3/library/dataclasses.html), a
feature of Python 3.7+ that is available on Python 3.6 too thanks to a
backward-compatible package.

This PR uses `@dataclass` to define the following classes:

- [X] `DetectorInfo` (previously `Detector`)
- [x] `FreqChannelInfo`
- [x] A way to quickly read from a TOML file/dictionary a list of detectors, possibly specifying entire channels at once
- [x] `InstrumentInfo`
- [x] Documentation
- [x] Agree on a few breaking changes for `Instrument`/`InstrumentInfo` (#65)


ziotom78:
The classes `DetectorInfo` and `FreqChannelInfo` are ready. This PR however should wait the resolution of the following minor issues in [litebird_imo](https://github.com/litebird/litebird_imo) before it can be merged:

- [Add parameter `sampling_rate_hz` to channel info](https://github.com/litebird/litebird_imo/issues/12)
- [Rename `pol_sensitivity_channel_uKarcmin` into `pol_sensitivity_channel_ukarcmin`](https://github.com/litebird/litebird_imo/issues/11)
- [Add measurement units for `bandcenter` and `bandwidth`](https://github.com/litebird/litebird_imo/issues/10)
- [The sampling rate should be restated in every detector's metadata](https://github.com/litebird/litebird_imo/issues/7)
- [Settle on a rule for uppercase/lowercase](https://github.com/litebird/litebird_imo/issues/4) (regarding `NET_ukrts`)

The reason why these are needed is because the fields in `DetectorInfo` and `FreqChannelInfo` refer to the new field names proposed in these issues.

ziotom78:
The PR seems in good shape now. Please have a look at the documentation here:

https://litebird-sim.readthedocs.io/en/dataclasses/detectors.html

It will be merged mid next week, if no concerns arise.

dpole:
Hi, I've got two comments. 

Starting with the quick one, it seems to me that we should change name to `detector.py`, even though I don't have a specific suggestion.

Second, we touched the topic elsewhere, but I promised during the last call that I'll drop a line also here. I'm am a bit afraid of the proliferation of classes storing the information of some part of the instrument. As I said in #53 , I prefer bare data: it is less safe, less OO (which is actually a pro to me) but more transparent and minimal (in the sense that it requires the user to get familiar with fewer concepts). I am fine with creating specialized classes for central pieces of information (as in this PR), but, looking more into the future, I am wondering whether we have clear ideas about where to draw the line and how to handle information beyond that line. Otherwise we'll end up with one dataclass per IMO entity.

ziotom78:
Hi @dpole , a few comments:

-   `detectors.py` was named similarly to `observations.py`, `simulations.py`, `quaternions.py`, `instruments.py`, i.e., plurals. If we decide to change it, we must change everything to stay consistent. Do you think it's worth doing this? As far as I can imagine, the user will just write `import litebird_sim`, and the name of the sub-modules will be rarely used.

-   I share your concerns about OO, but in my experience its biggest pain points are:

    -   Heavy use of complex inheritance hierarchies, which we avoid almost completely in our codebase

    -   Difficulty to access "private" members from the outside during debugging, which we avoid entirely because Python has no private members

    -   Having so many "getter" and "setter" methods in classes hurts eyes. However, so far we have avoided these entirely

    -   Boresome definition of constructors that have to initialize all the members one by one, which this PR solves

While I agree that OO can make the codebase more difficult to understand (the old joke about the [`NotificationStrategyFactory`](https://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html)!), I have never seen any data that backs up the claim that just using plain structs makes things more difficult for users. (Yes, I used the word "struct", because the way we are using classes in this codebase is more similar to C/Julia `struct` types than to C++ classes.) The ease of use depends on the API, not on the fact that we're avoiding the `class` keyword and the `object.member` notation (just *two* concepts to learn).

Finally, I too would like to avoid having too many classes, but my reason is against it is because of the "too many" part of the sentence, not because any OO construct must be avoided at all costs and in any case. I suggest we employ the following rule (abridged from [here](https://www.attrs.org/en/stable/why.html#dicts)):

1.    If the fields are always the same during runtime, implement a dataclass;
2.    If the fields can change at runtime depending on the user's input, use a dictionary.

The three classes `DetectorInfo`, `FreqChannelInfo`, and `InstrumentInfo` are meant to be a one-to-one replica of IMO objects, and thus they fall under point 1. For custom data like the `parameters` field of the `Simulation` class, we use the dictionary according to point 2.

dpole:
I see, thanks. I think we're talking about the same (or close enough) shade of gray from different angles. In particular, I agree that if we use dataclasses as C-like structures no complexity is added. This intention wasn't clear to me. A last question to clarify my practical concern about where to draw the line. Suppose that we add information about the HWP in the IMo with all sort of instrumental properties defined in it. How would you handle it? Is the idea to define a dedicated class?

About my first comment, it wasn't about the plural but rather the fact that there is more than detectors in the module (namely channel and instrument). 

Finally...I am a bit disappointed because I didn't get the `NotificationStrategyFactory` joke. It sounded like one of those nerdy things that make my day ;P

ziotom78:
> Suppose that we add information about the HWP in the IMo with all sort of instrumental properties defined in it. How would you handle it? Is the idea to define a dedicated class?

Yes, indeed: if the object is to be read from the IMO, it means that the set of metadata is "frozen" by a Specification Document, and thus it falls in the 1st category above.

In this way, we can get the information about an HWP using the construct

```python
hwp_info = HwpInfo.from_imo(imo, "/ref/to/hwp")
```

similarly to what we already do with `DetectorInfo.from_imo` and similar.

Now I get the point about `detectors.py`. What about something that underlines their close connection with the IMO? E.g., `imo_datatypes`.

dpole:
> Yes, indeed: if the object is to be read from the IMO, it means that the set of metadata is "frozen" by a Specification Document, and thus it falls in the 1st category above.

Maybe I'm overestimating the problem, but I imagine that the IMo will have a considerable number of different entities in the future. We are committing to writing a dataclass replicating the content of the imo specification for the entity. It seems to me an significant burden (maybe unnecessary) on developers.

Beyond the effort issue, maybe it is somehow forbidden by some IMo specification, but I can imagine an entity with many child entities. Walking this hierarchy of entities and calling at each step the proper constructor can lead to quite a bit of complicated code. Having an "anonymous" data container into which any entity may be loaded using a generic function seems a more straightforward and "scalable" solution to me.

ziotom78:
> Maybe I'm overestimating the problem, but I imagine that the IMo will have a considerable number of different entities in the future. We are committing to writing a dataclass replicating the content of the imo specification for the entity. It seems to me an significant burden (maybe unnecessary) on developers.

I do not expect that many details in the IMO will need to be accessed by our framework. (E.g., mechanical structure of the many hardware parts, thermal models, optical models, RF models, circuit diagrams, component datasheets, etc.) We're going to implement dataclasses only for those modules we actually need for the simulation.

> Beyond the effort issue, maybe it is somehow forbidden by some IMo specification, but I can imagine an entity with many child entities. Walking this hierarchy of entities and calling at each step the proper constructor can lead to quite a bit of complicated code. Having an "anonymous" data container into which any entity may be loaded using a generic function seems a more straightforward and "scalable" solution to me.

This framework is supposed only to load *data files* (not entities), and data files, like quantities, have no hierarchy in the current specification of the IMO.

PR #73, opened 2020-10-28
Bin map maker

dpole:
Purpose of this PR is to provide a minimal mapping tool, while more sophisticated procedures will be probably delegated to specialized libraries.

- [x] binning tool
- [x] diagonal pixel covariance
- [x] correction for the pixel covariance
- [ ] handle multiple maps
- [x] MPI parallel


dpole:
Distributed Observations are now handled. Note that currently all the detectors in an observation are mapped to the same map.

I think this is enough for a first merge, even though new functionalities may break the API. These may include

* mapping different set of detectors in the `Observation`s to different maps
* return also the covariance map

dpole:
Some tests have failed, but they are install issues. Shall we force merge?

ziotom78:
Merged! @dpole, might you please update the CHANGELOG with a reference to this PR?

dpole:
Sure. I could have done before merging

dpole:
Done

PR #74, opened 2020-10-31
Use caching in Travis builds and test Python 3.8 too

ziotom78:
This PR implements the following improvements:

- [x] Implement Travis CI caching to speed up builds on Linux machines
- [x] Test Python 3.8

Unfortunately, something seems to be broken under Mac OS X: builds that use any MPI library (OpenMPI/MPICH) break because `pip` installs packages using source tarballs instead of binary wheels (because it does not recognize the `mpicc` compiler) and this fails systematically. I had to disable Mac OS X builds running MPI, which means that now MPI is only tested under Linux. It seems that the latest XCode, released a few weeks ago, has introduced some breaking changes that affect a number of our dependencies.

I see two possibilities:

- We stop testing MPI under Mac OS X;
- Somebody with a Mac OS X (not me) debugs this.

Opinions, suggestions…?


ziotom78:
No comments so far… Can I assume that is ok to put MPI automatic testing aside under Mac OS X until we find a better solution?

mreineck:
I'd say yes. Probably not worth putting much more effort into CI tweaking until we know what we will use after Travis.

dpole:
Sorry for the slow feedback, I agree with Martin it is probably not worth investing time on travis

dpole:
Since we've moved to GHA (#77), I think this PR is not relevant anymore and can be closed. Caching in the CI is still relevant, even if not high priority

PR #76, opened 2020-11-11
map based simulations 

NicolettaK:
This PR implements the `Mbs `(Map based sims) class.

The class interfaces with the IMO, reads the channel infos and generates signal (CMB and foregrounds) and white noise maps accordingly, it can either write them to disk or return a dictionary that contains the generated maps. It makes use of `PySM3 `which is therefore added as a dependency in poetry.

# Examples

as usual a new `Simulation` object must be created 

```python

import litebird_sim as lbs
sim = lbs.Simulation(parameter_file='config.toml', name='test mbs')

```

The `Mbs` class can take information about the instrument in different ways:

1. in the standard mode it reads the channels for which to generate simulations in the configuration file, and interfaces with the IMO (see below an example of parameter file). In this case it should be initialized as follows:

```python
mbs = lbs.Mbs(sim)
```

2. A dictonary with specifications of a customized instrument can also be given, with the following keys:

```python
myinst = {}
myinst['ch1'] = {'freq':30, 'freq_band':12.,
                    'beam':5., 'P_sens':28.}
myinst['ch2'] = {'freq':300, 'freq_band':35.,
                    'beam':15., 'P_sens':45.}

mbs = lbs.Mbs(sim, instrument=myinst)

```

3. Finally a list of detectors or channels generated with the `DetectorInfo` or `FreqChannelInfo` can be used:

```python
    imo=sim.imo,
    url='/v1.0//satellite/LFT/L1-060/L00_003_QA_060B/detector_info',
)

mbs = lbs.Mbs(sim, detector_list=det)
```
or 

```python
freqch = lbs.FreqChannelInfo.from_imo(
    imo=sim.imo,
    url="/v1.0/satellite/LFT/L1-040/channel_info",
)

mbs = lbs.Mbs(sim, channel_list=freqch)

```

to then produce the simulated maps use the following line: 

```python
maps = mbs.run_all()
```

The maps can be saved on disk, if in the configuration file `save = true`. In this case all the simulated components are saved separately, and MonteCarlo realizations can be produced for noise or CMB. 

If `save = false` maps are not saved and the output would be a dictonary with one map (with the three IQU components) for each given channel or detector. In this case only one realization of CMB or noise can be simulated. 

Below an example of configuration file

```toml
[map_based_sims]

  [map_based_sims.instrument]
  IMO_version = 1.0
  telescopes = ['LFT', 'MFT', 'HFT']

  [map_based_sims.general]
  save = false
  nside = 16
  gaussian_smooth = true
  bandpass_int = true
  parallel_mc = true
  coadd = false

  [map_based_sims.noise]
  make_noise = true
  nmc_noise = 2
  seed_noise = 421431
  N_split = 2

  [map_based_sims.cmb]
  make_cmb = true
  cmb_ps_file = false
  cmb_r = 0
  nmc_cmb = 2
  seed_cmb = 4984

  [map_based_sims.fg]
  make_fg = true
  dust = 'pysm_dust_0'
  synch = 'pysm_synch_0'
  freefree =  'pysm_freefree_1'
  ame =  'pysm_ame_1'

  [map_based_sims.output]
  output_directory = 'test'
  output_string = 'example'
```

It would be great if you could have a look at it and provide comments.

Several things still need to be implemented:

  

- [ ] implement the report writing
- [x]   documentation


litebird_sim/mbs/mbs.py
```diff
@@ -0,0 +1,498 @@
...
+                        self.telescopes = config_inst['telescopes']
+                    except KeyError:
+                        try:
+                            self.channels = config_inst['channels']
+                        except KeyError:
+                                log.info("instrument dictonary should be pass to Mbs class")
+                if custom_instrument:
+                    if 'toml' in custom_instrument:
+                        self.instrument = toml.load(custom_instrument)
+                    elif 'npy' in custom_instrumet:
```
ziotom78, line 72: I believe there is a spelling error, it should be `custom_instrument`

litebird_sim/mbs/mbs.py
```diff
@@ -0,0 +1,498 @@
...
+        except KeyError:
+            if self.save==True:
+                self.coadd = False
+                log.info("setting coadd = False")
+            else:
+                self.coadd = True
+        try: self.parallel_mc = config_mbs['general']['parallel_mc']
+        except KeyError:
+            self.parallel_mc = False
+            log.info("setting parallel_mc = False")
```
ziotom78, line 129: The code above can be made much simpler by using a dedicated dataclass:

```python
from dataclasses import dataclass, fields
from typing import Union

# Define this *outside* the class `Mbs`
@dataclass
class MbsParameters:
    nside: int
    save: bool = True
    gaussian_smooth: bool = False
    bandpass_int: bool = False
    coadd: bool
    parallel_mc: bool = False
    make_noise: bool = False
    nmc_noise: int = 1
    seed_noise: Union[int, None] = None
    N_split: Union[int, bool] = False
    # etc.

    def __post_init__(self): # This is some magic enabled by the "dataclasses" method
        # Here you can manually fix parameters after they have been initialized
        # by a call to the constructor or to the `from_dict` static method
       if self.N_split == 1:
           self.N_split = False

       # Etc.

    @staticmethod
    def from_dict(dictionary):
        result = MbsParameters()
        for param in fields(result):
            if param.name in dictionary:
                setattr(result, param.name, dictionary[param.name])

        # This is not a constructor, so we need to call it manually
        result.__post_init__()
        return result
```

This is the same approach used in the class `DetectorInfo` (file `litebird_sim/detectors.py`), and it has the following advantages:

1.   All the parameters used by the simulation are listed in one place, and their default values are stated explicitly and collected within a narrow piece of code.

2.   The user can get a list of the parameters accepted by the module, as well as their default values, writing

          ?MbsParameters

      at the IPython prompt.

2.   There is no need to manually read them one by one from a dictionary, as the `fields` method permits to iterate over them; in fact, the method `check_parameters` becomes:

      ```python
      def check_parameters(self):
          self.config = MbsParameters(self.parameters)
          # That's it! You can refer to any parameter later as `self.config.units`,
          # `self.config.parallel_mc`, etc.
      ```

      In fact, the method `check_parameters` becomes so trivial that it could probably be moved in the `Mbs.__init__` method.

3.   An object `MbsParameters` can be easily created on-the-fly, and this is probably useful in the unit tests for the class.

litebird_sim/mbs/mbs.py
```diff
@@ -0,0 +1,498 @@
...
+                    if (self.save) and not (os.path.exists(out_dir+nmc_str)):
+                            os.makedirs(out_dir+nmc_str)
+                    if N_split:
+                        split_rms = tot_rms*np.sqrt(N_split)
+                        noise_map = np.zeros((3, npix))
+                        for hm in range(N_split):
+                            noise_map_split = np.random.randn(3, npix)*split_rms
+                            noise_map += noise_map_split
+                            file_name = f'{chnl}_noise_SPLIT_{hm+1}of{N_split}_{nmc_str}_{file_str}.fits'
+                            file_tot_path = f'{out_dir}{nmc_str}/{file_name}'
```
ziotom78, line 247: It's better to make the code OS-agnostic; the slash character `/` does not work under Windows. Use the `Path` object from the `pathlib` library:

```python
from pathlib import Path

# ...

nmc_dir = Path(f'{out_dir}{nmc_str}')

# The operator '/' is defined on paths and it concatenates two paths in
# an OS-independent way. Thus, it will be expanded to '/' under Unix
# systems, and to `\` under Windows
file_tot_path = nmc_dir / file_name
```

litebird_sim/mbs/mbs.py
```diff
@@ -0,0 +1,498 @@
...
+                            lbs.write_healpix_map_to_file(file_tot_path, noise_map_split, column_units=col_units)
+                        noise_map = noise_map/N_split
+                    else:
+                        noise_map = np.random.randn(3, npix)*tot_rms
+                    if self.save:
+                        file_name = f'{chnl}_noise_FULL_{nmc_str}_{file_str}.fits'
+                        file_tot_path = f'{out_dir}{nmc_str}/{file_name}'
```
ziotom78, line 254: See above, it's better to avoid using `/` explicitly.

litebird_sim/mbs/mbs.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def generate_cmb(self):
+        instr = self.instrument
+        nmc_cmb = self.nmc_cmb
+        nside = self.nside
+        npix = hp.nside2npix(nside)
+        smooth = self.gaussian_smooth
+        parallel = self.parallel_mc
+        root_dir = self.out_dir
+        out_dir = f'{root_dir}/cmb/'
```
ziotom78, line 271: Same as above

litebird_sim/mbs/mbs.py
```diff
@@ -0,0 +1,498 @@
...
+            rank = comm.Get_rank()
+            size = comm.Get_size()
+        if not os.path.exists(out_dir) and rank==0:
+            os.makedirs(out_dir)
+        if cmb_ps_file:
+            cl_cmb = hp.read_cl(cmb_ps_file)
+        else:
+            cmb_ps_scalar_file = os.path.join(
+                os.path.dirname(__file__),
+                '../datautils/Cls_Planck2018_for_PTEP_2020_r0.fits')
```
ziotom78, line 291: Same as above

litebird_sim/mbs/mbs.py
```diff
@@ -0,0 +1,498 @@
...
+        fg_models = self.fg_models
+        components = self.fg_components
+        ncomp = len(components)
+        rank = 0
+        col_units = [self.units, self.units, self.units]
+        if parallel:
+            comm = lbs.MPI_COMM_WORLD
+            rank = comm.Get_rank()
+        if not os.path.exists(out_dir) and rank==0 and self.save:
+            os.makedirs(out_dir)
```
ziotom78, line 368: These two lines can be substituted by this single call:

```python
Path(out_dir).mkdir(parents=True, exists_ok=True)
```

litebird_sim/mbs/mbs.py
```diff
@@ -0,0 +1,498 @@
...
+        if rank==0:
+            if not self.save:
+                dict_fg = {}
+            for cmp in components:
+                if not os.path.exists(out_dir+cmp) and rank==0 and self.save:
+                    os.makedirs(out_dir+cmp)
```
ziotom78, line 374: Same as above

ziotom78 (commented):
It's better to remove the `.DS_Store` file from the commit.

ziotom78:
Ok, I have tried to create some documentation for the module:

- Docstrings
- A chapter in the User's manual

I think this can be merged now. @NicolettaK , what do you think?

NicolettaK:
Thanks a lot @ziotom78 for this! We should merge the PR. 
I'll add few details in the documentation in the next days.

PR #77, opened 2020-11-26
Move the CI from Travis to GitHub Actions

dpole:
I think we are all set. 
It took longer than I thought, but overall I would say that GHA is a pretty neat and powerful solution for our needs -- more than Travis in my opinion.

Note that the test matrix does not include windows because this PR only aims at replacing the role of `.travis.yml`.

I also took the opportunity to do some clean-up from all the Travis-specific directives. @ziotom78 could you please give it a look and check that I didn't leave out something we should keep?

Also, testing all the combinations in strategy matrix might be an overkill. Let me know what you think. In any case, the total runtime (without any caching) is about 15min, which sounds reasonable to me.


mreineck:
Great!
`litebird/ducc` is now also almost fully migrated; I think I even got the native Windows compilation to work!

ziotom78:
Great job, thanks!

> Also, testing all the combinations in strategy matrix might be an overkill. Let me know what you think. In any case, the total runtime (without any caching) is about 15min, which sounds reasonable to me.

Agreed. I would just test the oldest and the newest-but-not-so-new Python versions (currently 3.6 and 3.8) on Linux, and just the latter on Mac (3.8). The idea is that people running Mac are usually willing to keep their machine updated, because of Apple's policy for planned obsolescence :-)

dpole:
> Agreed. I would just test the oldest and the newest-but-not-so-new Python versions (currently 3.6 and 3.8) on Linux, and just the latter on Mac (3.8). The idea is that people running Mac are usually willing to keep their machine updated, because of Apple's policy for planned obsolescence :-)

I guess it was only a half joke, so I took it seriously and now only 3.6 is tested for macos ;)

dpole:
I think we can merge, please let me know if anybody thinks there is something we should check further

ziotom78:
It looks good! Please merge it as soon as you wish.

PR #95, opened 2021-01-21
Update pyproject.toml

mreineck:
I'm happy with being listed, but strictly speaking I have not yet contributed any code to this repository, so it's difficult to say whether I should appear in there.


ziotom78:
Hi @mreineck , thanks!

> strictly speaking I have not yet contributed any code to this repository, so it's difficult to say whether I should appear in there.

The point is that `litebird_sim` benefits from other repositories as well, either by importing them directly (like `ducc0`) or by reusing their code indirectly (like `toast-litebird`), and thus it's fine to be listed here as well.

PR #96, opened 2021-01-22
Add support for Singularity

ziotom78:
This PR adds a few scripts that ease the build of Singularity files.
It includes a few screencasts and a new section in the «Installation»
chapter of the User's Manual. Singularity could be a great tool to
ease the installation of the framework on HPC clusters.


mreineck:
Sorry if this is an uninformed question, but does the singularity container contain a compiled vesion of `ducc`? If yes, then we have to make sure it will run on all CPUs of a given class, i.e. we need to disable native optimization. This is certainly OK for users trying the package at home, but not really in an HPC environmant.

ziotom78:
> Sorry if this is an uninformed question, but does the singularity container contain a compiled vesion of ducc?

Yes, it does. I do not think it is going to be a problem if users create the container on the same HPC cluster they are going to use it, right?

mreineck:
Yes, that's correct. I probably misunderstood our intention, because I thought we wanted to save HPC users the trouble of building the package altogether; but I expect now that Sinularity's advantage is rather to provide a uniform environment where building and installing works more smoothly than on inhomogeneous "bare" HPC environments.

ziotom78:
Yep, that's right. I was triggered to do this when I was told that there were problems building the package because of conflicts between an Anaconda-based environment and the system-wide Python distribution on a HPC cluster used by somebody in the our collaboration. Hope this will help.

ziotom78:
I believe this PR is complete now. Feel free to comment!

PR #100, opened 2021-02-05
Noise generation

unfunfunt:
What do people think about the interface between the Simulation Class and all the specific classes that generate particular effects (like cosmic rays, adc issues, noise, etc.)? Noise is the first of those that has been implemented, as far as I can tell, so I thought it was worth discussing because other people will model their interfaces on this one. 

One option would be adding noise like this:
sim.add_noise(observation, 'white')
The advantage of this is that the end user doesn't have to know where the particular thing they want to simulate lives, everything is just a member of sim. The downside is that simulation.py will contain every interface function to every simulated effect, and that isn't super modular.

We could also do something like this:
lbs.noise.add_noise(observation, 'white')
This means people need to know which subfile the effect they want to simulate lives in, but it solves the problem of simulation.py getting too large. 

Finally, we could make generator classes to the effects members of sim, like so:
sim.noise.add_noise(observation, 'white')
To me this feels like the worst of the three options, as it still requires changes in simulation.py for each effect, and it requires people to know what class it is found it. 

Please comment if you have any options.


dpole:
I vote for the second. Actually I vote for `lbs.noise.add_white_noise(observation)` ;)

ziotom78:
Hi @unfunfunt , thanks for the PR! I agree with @dpole ; the potential discoverability issue should be mitigated if we add a chapter about «noise generation» to the user's manual.

unfunfunt:
This is my latest version, which is close to working well enough to be used in a demo next week. A couple things I changed in the rest of the code, feel free to complain about them:
1 - I changed newly allocated TOD from np.empty to np.zeros, because we decided that tod operations should be additive. Thus, you can allocate a TOD and then add something to it without the need for an intermediate step of setting it to 0. If there was another way this should have been done, please let me know.
2 - I added support for integer arguments to duration_s in Simulation(). This is a usage case I think we should support, but if people have strong opinions please comment.

I also have a question about the framework: What is the basic unit that the TOD should be stored in? The noise is currently in units of uK (I think although someone should check that they agree). Is this what we will be using in the pipeline, or will it be in K, or some uncalibrated detector unit? I think it's good to standardize something now before people start writing operators that use different conventions. 

Please look over the code and see what you think. If we want to demo this next week, we should probably start the review process soon.

litebird_sim/imo/imo.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, flatfile_location=None, url=None, user=None, password=None):
                 location = config["repositories"][0]["location"]
                 self.imoobject = ImoFlatFile(location)
             except FileNotFoundError:
-                log.warning('IMO config file "%s" not found', str(CONFIG_FILE_PATH))
+                log.warning('IMO config file "%s" not found. Have you run the initial setup procedure (python -m litebird_sim.install_imo)?', str(CONFIG_FILE_PATH))
```
ziotom78, line 28: Nice touch, thanks!

litebird_sim/noise.py
```diff
@@ -0,0 +1,87 @@
...
+            for i in range(ob.tod.shape[0]):
+                if noisetype == 'white':
+                    generate_white_noise(ob.tod[i][:], ob.net_ukrts[i]*np.sqrt(ob.sampling_rate_hz))
+                elif noisetype == 'one_over_f':
+                    generate_one_over_f_noise(ob.tod[i][:], ob.fknee_mhz[i], ob.alpha[i], ob.net_ukrts[i]*np.sqrt(ob.sampling_rate_hz), ob.sampling_rate_hz)
+
+        else:
+            raise TypeError("Array with shape " + ob.tod.shape + " not supported in add_noise. Try a 1 or 2 D array instead.")
+
+def generate_white_noise(data, sigma):
```
ziotom78, line 40: Might you please change the name of the parameter "sigma" so that the proper measurement unit is appended? Like `sigma_uk`, `sigma_uk2`, etc.

litebird_sim/noise.py
```diff
@@ -0,0 +1,87 @@
...
+    '''Adds white noise with the given sigma to the array data
+    To be called from add_noise.
+
+    data: 1-D numpy array
+    sigma: the varience of the desired white noise
+
+    '''
+    data += np.random.default_rng().normal(0, sigma, data.shape)
+
+def generate_one_over_f_noise(data, fknee, alpha, sigma0, freq):
```
ziotom78, line 50: Same as above: `fknee_hz`, `sigma0_uk`, `freq_hz` (if these are the correct units).

litebird_sim/observations.py
```diff
@@ -96,7 +96,7 @@ def __init__(
             self._n_blocks_time = 1
 
         if allocate_tod:
-            self.tod = np.empty(
+            self.tod = np.zeros(
```
ziotom78, line 99: I like this change: it makes everything simpler from the point of view of the user.

litebird_sim/simulations.py
```diff
@@ -612,8 +613,9 @@ def create_observations(
         assert (
             self.start_time is not None
         ), "you must set start_time when creating the Simulation object"
+
         assert isinstance(
-            self.duration_s, float
+            self.duration_s, (float, int)
```
ziotom78, line 618: I can't tell how many times I wrongly passed an integer to this parameter… Thanks, I fully approve this addition.

ziotom78 (commented):
Thanks a lot for the PR, @unfunfunt ! I believe the implementation is almost perfect; I've added a few comments below.

Two general points:

-   I would like to merge this only once we have some documentation in the User's manual. I would add a short chapter named *Noise generation*, or something like it. The source code (in RST format) for the manual is in `docs/src`.

-   This is the first time we inject something in the TOD, and you have raised the issue of the measurement unit. I would advocate for K, as I prefer to use SI units. (The reason why we use uKrts for noise is because this is what the IMO uses). Any other opinion on this?

ziotom78:
Hi @unfunfunt , I have checked the errors in the GitHub Action build, and it seems that they are happening because the build machines do not have a working IMO installed, and thus the UUIDs are not recognized.

We cannot provide a JSON file with the nominal IMO in the repository for obvious reasons, so the solution we have adopted so far is to directly create `DetectorInfo` objects in the documentation. Thus, instead of

```python
det = lbs.DetectorInfo.from_imo(sim.imo, '8b20679d-f30e-4d17-8b72-276e926fea73')
```

you should use the following style in the examples provided with the documentation:

```python
det = lbs.DetectorInfo(
    alpha=1.0,
    fknee_mhz=10.0,
    # Et cetera
)
```

Be sure to use round numbers, so that it is clear that they are just mock values and not coming from an official specification.

unfunfunt:
At this point I am stuck. I can't run all of run_test.sh locally because it fails with 
```Extension error:
Could not import extension sphinxcontrib.bibtex (exception: No module named 'sphinxcontrib.bibtex')
```
I don't know enough about sphinx documentation to know what the issue in the build system is, which seems to be related to my piece of test code where I have printed the output of the noise generator. I modelled this off of detectors.rst, but I'm not sure what I have done wrong here that isn't in that code

ziotom78:
> At this point I am stuck. I can't run all of run_test.sh locally because it fails with

That's strange, `sphinxcontrib-bibtex` is a dependency listed in `pyproject.toml`. Are you using Poetry? If you run `poetry show`, is `sphinxcontrib-bibtex` in the list?

Regarding the failing tests, you forgot to add the expected output in a `testoutput::` section. Have a look at [`scanning.rst`](https://raw.githubusercontent.com/litebird/litebird_sim/master/docs/source/scanning.rst). (Perhaps you should reduce the amount of data echoed through `print`, to make the output easy to read: everything included in `testoutput::` will appear in the documentation website.)

unfunfunt:
Poetry show lists sphinxcontrib-bibtex, although it is coloured red instead of green so I'm not sure what that means. 
I think the problem might be arising because the noise relies on a random number generator, so I cannot provide test output that is accurate every time, as each noise realization will be different. I will just cut out the print entirely and see if that passes.

ziotom78:
> I think the problem might be arising because the noise relies on a random number generator, so I cannot provide test output that is accurate every time, as each noise realization will be different.

I believe the best way to go is to enable the possibility to provide a custom number generator, and initialize it with a fixed seed (e.g., `seed=1234`). In this way, the numbers would be deterministic.

This option might be handy when we will add MPI support to the noise generation module, as there are generators that are handy in concurrent codes and the user might want to tune them.

mreineck:
Just wanted to add that `numpy.random`  has very convenient ways to generate reproducible random numbers, even in the presence of MPI; see for example the documentation for `numpy.random.SeedSequence`. This was added a few years ago.

unfunfunt:
This latest push looks like it failed for reasons unrelated to the code: 
`Run ./bin/install-mpi.sh openmpi
+ echo Installing openmpi
Installing openmpi
+ sudo apt-get install -y -q openmpi-bin libopenmpi-dev
Reading package lists...
Building dependency tree...
Reading state information...
The following additional packages will be installed:
  ibverbs-providers libcaf-openmpi-3 libcoarrays-dev libcoarrays-openmpi-dev
  libevent-dev libevent-extra-2.1-7 libevent-openssl-2.1-7 libfabric1
  libhwloc-dev libhwloc-plugins libhwloc15 libibverbs-dev libibverbs1
  libnl-3-dev libnl-route-3-200 libnl-route-3-dev libnuma-dev libopenmpi3
  libpmix2 libpsm-infinipath1 libpsm2-2 librdmacm1 libxnvctrl0
  ocl-icd-libopencl1 openmpi-common
Suggested packages:
  libhwloc-contrib-plugins openmpi-doc opencl-icd
The following NEW packages will be installed:
  ibverbs-providers libcaf-openmpi-3 libcoarrays-dev libcoarrays-openmpi-dev
  libevent-dev libevent-extra-2.1-7 libevent-openssl-2.1-7 libfabric1
  libhwloc-dev libhwloc-plugins libhwloc15 libibverbs-dev libibverbs1
  libnl-3-dev libnl-route-3-200 libnl-route-3-dev libnuma-dev libopenmpi-dev
  libopenmpi3 libpmix2 libpsm-infinipath1 libpsm2-2 librdmacm1 libxnvctrl0
  ocl-icd-libopencl1 openmpi-bin openmpi-common
0 upgraded, 27 newly installed, 0 to remove and 4 not upgraded.
Need to get 6243 kB of archives.
After this operation, 26.5 MB of additional disk space will be used.
Get:1 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libnl-route-3-200 amd64 3.4.0-1 [149 kB]
Get:2 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libibverbs1 amd64 28.0-1ubuntu1 [53.6 kB]
Get:3 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 ibverbs-providers amd64 28.0-1ubuntu1 [232 kB]
Get:4 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libpsm-infinipath1 amd64 3.3+20.604758e7-6 [168 kB]
Get:5 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libpsm2-2 amd64 11.2.86-1 [178 kB]
Get:6 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 librdmacm1 amd64 28.0-1ubuntu1 [64.9 kB]
Get:7 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libfabric1 amd64 1.6.2-3 [396 kB]
Get:8 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libhwloc15 amd64 2.1.0+dfsg-4 [134 kB]
Err:9 http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 libxnvctrl0 amd64 440.82-0ubuntu0.20.04.1
  404  Not Found [IP: 52.252.75.106 80]
Get:10 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 ocl-icd-libopencl1 amd64 2.2.11-1ubuntu1 [30.3 kB]
Get:11 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libhwloc-plugins amd64 2.1.0+dfsg-4 [14.4 kB]
Get:12 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libpmix2 amd64 3.1.5-1 [442 kB]
Get:13 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libopenmpi3 amd64 4.0.3-0ubuntu1 [1978 kB]
Get:14 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libcaf-openmpi-3 amd64 2.8.0-1 [35.5 kB]
Get:15 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libcoarrays-dev amd64 2.8.0-1 [28.2 kB]
Get:16 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 openmpi-common all 4.0.3-0ubuntu1 [151 kB]
Get:17 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 openmpi-bin amd64 4.0.3-0ubuntu1 [67.4 kB]
Get:18 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libcoarrays-openmpi-dev amd64 2.8.0-1 [34.2 kB]
Get:19 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libevent-extra-2.1-7 amd64 2.1.11-stable-1 [60.0 kB]
Get:20 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libevent-openssl-2.1-7 amd64 2.1.11-stable-1 [14.3 kB]
Get:21 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libevent-dev amd64 2.1.11-stable-1 [261 kB]
Get:22 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libnl-3-dev amd64 3.4.0-1 [92.2 kB]
Get:23 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libnl-route-3-dev amd64 3.4.0-1 [166 kB]
Get:24 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libnuma-dev amd64 2.0.12-1 [32.4 kB]
Get:25 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libhwloc-dev amd64 2.1.0+dfsg-4 [205 kB]
Get:26 http://azure.archive.ubuntu.com/ubuntu focal/main amd64 libibverbs-dev amd64 28.0-1ubuntu1 [444 kB]
Get:27 http://azure.archive.ubuntu.com/ubuntu focal/universe amd64 libopenmpi-dev amd64 4.0.3-0ubuntu1 [798 kB]
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/n/nvidia-settings/libxnvctrl0_440.82-0ubuntu0.20.04.1_amd64.deb  404  Not Found [IP: 52.252.75.106 80]
Fetched 6232 kB in 0s (45.1 MB/s)
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.`

ziotom78:
> This latest push looks like it failed for reasons unrelated to the code

I see, it seems that some servers are currently down. I re-triggered the build a couple of times with no effect. Let's wait a couple of days, usually they don't take long to fix this kind of problems.

unfunfunt:
Sounds good!

PR #103, opened 2021-02-09
Add a tool to browse the IMO

ziotom78:
This PR implements a simple browser for the IMO. It is meant to provide a quick way to copy paths to objects and UUIDs to the clipboard, and it is *not* meant to be an alternative to the full web-interface to the IMO that is being developed.

I decided to implement a text-only interface, so that the tool can be used seamlessy even when connected through `ssh` to a remote machine. The curses interface uses [Asciimatics](https://github.com/peterbrittain/asciimatics), which supports Linux, Mac OS X, and Windows.

Here are the features that are currently implemented:

- [X] Basic text-mode interface implemented using [Asciimatics](https://github.com/peterbrittain/asciimatics)
- [X] Ability to copy paths and UUIDs to the clipboard using [Pyperclip](https://github.com/asweigart/pyperclip)
- [X] Add a section about the tool in the User's manual


ziotom78:
Here is a screenshot:

imobrowser_demo

Note the presence of the «Copy path» and «Copy UUID»: they work on the data file that is currently selected.

I plan to merge this PR soon, as the feature it adds does not clash with anything else in the framework.

Issue #109, opened 2021-03-25
Use MPI with create_observations


paganol:
Hi there, I'm trying to use mpi with sim.create_observations. I cannot understand how to pass correctly the communicator. Setting it in the initialization of lbs.Simulation is not working. Am I supposed to initialize obs only with lbs.Observation for handling mpi runs? It should not be the case, right?
Another small thing, reading the documentation, lbs.Simulation should have the flag use_mpi, but it doesn't exist anymore.

dpole:
Hi Luca, yes, everything should be set for you when you construct the simulation with an MPI communicator. Note that you have to set ``distribute=False``. I'd remove this flag altogether, I find it misleading.
About the doc, I believe you are right and it should be fixed. @ziotom78 do you agree?

ziotom78:
@dpole, yes, I agree both with the removal of `distribute` and of `use_mpi`. I'll do it this morning

paganol:
Ok, thank you both

PR #110, opened 2021-04-07
Improve MPI-related and Simulation docs

ziotom78:
This PR adds the following features to the documentation:

-   It corrects the syntax used to turn MPI on/off in a `Simulation` object (we do no longer use the `use_mpi` flag);
-   It adds a (long) docstring to `Simulation.create_observations`;
-   It removes references to `use_mpi` that were still present in docstrings;
-   It fixes a few minor errors in the Sphinx documentation.


ziotom78:
This should fix issue #109. Note that I decided *not* to remove `distribute`, as I realized that there are significant use cases where it might be indispensable (e.g., when you do not want to rely on the standard way of distributing observations offered by the framework).

dpole:
Thanks Maurizio, I am still a bit confused by all the possible interactions that can happen between the parameters `num_of_obs_per_detector`, `distribute` and `n_blocks_time`.

`num_of_obs_per_detector` chunks the time and keep the chunks as a list of observation. How the content in the memory of the process changes depending on the values of `distribute` or `Simulation.mpi_comm`?

ziotom78:
I have renamed the confusingly-named `distribute` parameter to `split_list_over_processes` and have added a (quite long) docstring to `Simulation.create_observations`. @dpole , @paganol , this should answer the perplexities we had, what do you think?

PR #122, opened 2021-07-06
Compute the position and velocity of the spacecraft

ziotom78:
When finished, this PR will implement a module that will simulate the orbit of the LiteBIRD spacecraft around the L2 point. It will be used to simulate the observability of Solar System objects as well as the computation of the solar/orbital dipole.

- [X] Define `DEFAULT_COORDINATE_SYSTEM` in a separate module (`coordinates`), so that it can be referenced every time we use `astropy.coordinates`;
- [X] Make `litebird_sim/scanning.py` use `DEFAULT_COORDINATE_SYSTEM` instead of hardwiring the coordinate system in the code
- [X] Implement `compute_l2_pos_and_vel` (compute the location and the velocity of the L2 point)
- [X] Implement `compute_lissajous_pos_and_vel` (compute the location and the velocity of the spacecraft wrt the L2 point)
- [X] Add the `SpacecraftOrbit` dataclass
- [x] Harmonize the `compute_l2_pos_and_vel` and `compute_lissajous_pos_and_vel` with the `Observation` class, so that the caller can know the position and velocity of the spacecraft at the time of an observation
- [x] Implement tests
- [x] Update the documentation


mreineck:
I have no experience with numba so far, so maybe this is a stupid question about the dipole calculation: does the level of approximation really make a significant difference as far as performance is concerned?
If this is, say, a ten percent effect, I'd propose to go for code simplification and just implement the exact dipole contribution.

ziotom78:
> I have no experience with numba so far, so maybe this is a stupid question about the dipole calculation: does the level of approximation really make a significant difference as far as performance is concerned?
> If this is, say, a ten percent effect, I'd propose to go for code simplification and just implement the exact dipole contribution.

Hi Martin, I'll give my answer and then let @paganol to comment.

I agree that there should be no appreciable difference in the running time of the many implementations of the dipole estimator. However, if we want to do comparisons with results produced by other existing codes, there are chances that those codes use different conventions to compute the dipole. Having a reasonable set of models implemented in our framework could help people in doing meaningful comparisons with other estimates.

Moreover, it might be interesting to compare the estimates between a rough model and a more refined model. If we discover that some correction is negligible, then this might be a hint for theoreticians working on LiteBIRD that it's ok to use a more simplified formula in their calculations.

paganol:
Hi @mreineck, I agree with @ziotom78's comment. Different approximations have probably similar running time. I think the reason for having them implemented is in the possibility of testing the impact of higher-order corrections, like frequency-dependent kinematic quadrupole. For example, those terms might have an impact on calibration and, with this implementation, we can test it.

mreineck:
Thanks a lot for the explanations!

PR #123, opened 2021-07-08
mbs can now handle units

NicolettaK:
This PR modifies the `Mbs` module in order to:

- handle output units
- make K_CMB as default units
- do not generate FG maps when `make_fg = False`,  even if `fg_model` is defined


paganol:
Ciao @NicolettaK, can we merge this?

ziotom78:
Hi @paganol, the code should be ok, but I waited before merging it because the modifications were not documented. It should be trivial to do so (just add a couple of sentences), I'll have a look.

ziotom78:
I have had a look at the patch and decided that the change in the docstring should be enough. It would have been nice to extend a bit the short example at the beginning of chapter «Synthetic sky maps» in the User's Manual, but for the moment let's be satisfied with this.

Issue #126, opened 2021-09-09
Failing tests are not signalled by GitHub


ziotom78:
Old test runs like [this](https://github.com/litebird/litebird_sim/runs/3373029359) show that failing tests are not correctly reported as such in GitHub Actions, which mistakenly marks them with «success»:

```
…
test/test_destriper.py::test_destriper FAILED                            [  1%]
test/test_detectors.py::test_detector_from_dict PASSED                   [  3%]
test/test_detectors.py::test_detector_from_toml PASSED                   [  5%]
test/test_detectors.py::test_detector_from_imo PASSED                    [  7%]
…
```

It seems that the problem is a lack of checks in `tests/run_tests.sh` (which is used by the test workflow).

PR #130, opened 2021-09-10
Install pre-commit and fix the configuration for flake8/black

ziotom78:
This small PR installs [`pre-commit`](https://pre-commit.com/) and documents its usage. This module is optional for developers, but it can save a lot of time when doing modifications to the repository. Now that issue #126 has been fixed, all the tests in `bin/run_tests.sh` are ran and checked properly by GitHub, and things are more likely to go bad! The script checks for proper code formatting and a number of stylistic errors (e.g., too long lines, unused variables).

The usage of `pre-commit` enables developers to discover these problems *before* running `git push`, as `black` and `flake8` are ran automatically whenever the user calls `git commit`.

There is no need for developers to use this facility, so if you're happy with your workflow, this PR does not affect it.


ziotom78:
I have removed the AppVeyor config file, which was used only to check that the code was runnable under WSL but it has not been maintained since ages. We'll decide if we want to add tests for Windows using GitHub Actions, which would be far easier to handle.

PR #131, opened 2021-09-18
Module for scanning a map and filling TOD 

paganol:
This PR provides a module able to fill obs.tod given pointing information and input map provided by mbs. It also makes obs compatible with `mapping.py` optionally filling `obs.psi` and `obs.pixind`.
For now, no ecliptic-to-galactic coordinate rotation is implemented internally. So `maps` should be rotated to ecliptic before being fed to the function `scan_map_in_observations`. Probably having an option in mbs for dealing with that is worthed, I'll open an issue for this.
The memory handling can be improved, in particular, if psi and pixind are not filled. 

Some points are still open:
- [ ] Check numba implementation
- [ ] Dealing with e2g rotation (if necessary)
- [ ] Move ang2pix in `scan_map_for_one_detector` to save some memory?
- [ ] Implement test
- [ ] Update documentation

### DIFF
--- a/docs/source/sky_maps.rst
+++ b/docs/source/sky_maps.rst
@@ -1,3 +1,5 @@
+.. _Mbs:
+
 Synthetic sky maps
 ==================
 
--- a/docs/source/timeordered.rst
+++ b/docs/source/timeordered.rst
@@ -80,20 +80,20 @@ shows:
    lbs.scan_map_in_observations(obs, pointings, hwp_radpsec, in_map)
    
    for i in range(obs.n_samples):
-       print(f"{obs.tod[0][i]:.5e}")
+       print(f"{obs.tod[0][i]:.5f}")
 
 .. testoutput::
 
-   -3.89414e-07
-   -1.44749e-01
-   -2.61039e-01
-   -3.45981e-01
-   -3.97462e-01
-   -4.14204e-01
-   -3.95788e-01
-   -3.42674e-01
-   -2.56182e-01
-   -1.38462e-01
+   0.00000
+   -0.14475
+   -0.26104
+   -0.34598
+   -0.39746
+   -0.41420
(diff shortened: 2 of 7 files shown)


ziotom78 (approved):
Very nice work! It looks good to me, we can merge it immediately.

ziotom78:
Ooops, I see that the new test in the documentation (file `timeordered.rst`) fails because of some rounding error:

```
Expected:
    -3.89414e-07
    -1.44749e-01
    -2.61039e-01
    -3.45981e-01
    -3.97462e-01
    -4.14204e-01
    -3.95788e-01
    -3.42674e-01
    -2.56182e-01
    -1.38462e-01
Got:
    8.74228e-08
    -1.44749e-01
    -2.61039e-01
    -3.45981e-01
    -3.97462e-01
    -4.14204e-01
    -3.95788e-01
    -3.42674e-01
    -2.56182e-01
    -1.38462e-01
```

Probably the best way to fix this is to change the format of the output numbers from %.5e to %.5f, so that the first number is rounded to zero.

ziotom78:
Regarding the point about `ang2pix`, I would merge this PR as is and open a new issue for it.

Should we decide that it is worth to save some memory, I wrote an implementation of `ang2pix` in Numba a couple of years ago in [this gist](https://gist.github.com/ziotom78/40abc5a780c6979b19fef3b82054eb83). It just provides `ang2pix` for RING maps and a set of test cases. It's faster than Healpy but slower than DUCC; however, the fact that the function is written in Numba makes it optimal if called within a `for` loop in another Numba function, as it is the case here. (Numba does not makes calls to C functions as efficient as calls to Numba functions, as the latter can be inlined and included in the LLVM global optimization ran by its JIT compiler.) We might include those lines of code and rewrite the corresponding loop.

Issue #132, opened 2021-09-18
Option for outputting maps in ecliptic coordinates in Mbs


paganol:
We already mention something similar in #116. But, even if at some point we'll deal with coordinates in `scan_map.py`, have the possibility of outputting maps in ecliptic might be a nice feature of Mbs. I can take care of this, @NicolettaK, probably I'll just need your feedback at some point.

PR #133, opened 2021-09-19
Option in Mbs for outputting maps in ecliptic coordinates

paganol:
This quickly addresses #132. Only the dictionary returned is affected by this change, all the other features are untouched


ziotom78:
Hi @paganol, there is a conflict with `master` because of the default units for Mbs, which are now K_CMB. Is it fine to keep that default? (I believe so.)

paganol:
Can I merge this?

ziotom78:
Yeah, sure, if the conflict is resolved we can merge this immediately. (Please update the CHANGELOG when you do this.)

PR #138, opened 2021-10-08
Functions to write TODs to disk

ziotom78:
- [X] Add dependency on h5py
- [X] Use a more precise type for Observation.__init__
- [X] Add code to write TODs to HDF5 files
- [ ] Add tests
- [ ] Add documentation


ziotom78:
Spurious commits were included in this PR, better to start from scratch…

PR #147, opened 2021-11-26
Use a cache to store PySM3 data files

ziotom78:
As it is often the case that PySM3 data files are not available due to network outage, this PR uses https://github.com/actions/cache to cache it. It should be quite efficient, as cache files are compressed using [Zstandard](http://facebook.github.io/zstd/), which in my tests showed very good compression ratios for PySM3 maps.

I use here the trick explained in the [PySM3 User's Manual](https://pysm3.readthedocs.io/en/latest/models.html#input-templates)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,12 +27,15 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
 
+      # Unfortunately we must use different keys to store the same data,
+      # otherwise GitHub will complain when multiple jobs are going to
+      # store it
       - name: Cache PySM3 data files
         id: cache-pysm3
         uses: actions/cache@v2
         with:
           path: pysm3-data
-          key: pysm3-data
+          key: pysm3-data-${{ runner.os }}-${{ matrix.python }}-${{ mpi }}
 
       - name: Register PySM3 data path
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,6 +60,13 @@ jobs:
         env:
           DUCC0_OPTIMIZATION: none
 
+      - name: Enable rich traceback
+        run: |
+          export DESTPATH="$(poetry env info -p)/lib/python${{ matrix.python }}/site-packages/sitecustomize.py"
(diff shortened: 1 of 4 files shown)


ziotom78:
Not relevant for the PR, but just to keep a record of my tests and to keep in mind the amount of storage we can save with different compression schemes. It's also a testimony to Zstandard's awesomeness!

I downloaded the PySM3 data archive and create a tarball:

```
$ git clone https://github.com/galsci/pysm-data pysm3-data
$ tar cf archive.tar pysm3-data
```

and then I compressed `archive.tar` using `gzip`, `bzip2`, and `zstd`. Here are the file sizes:

| File | Size |
|-|-|
| Uncompressed | 754M |
| BZip2 | 641M |
| Gzip | 632M |
| Zstandard | 615M |

image

Zstandard is the winner here. However, the most impressive result comes from compression speed:

| File | Compression time |
|-|-|
| BZip2 | 100 s |
| Gzip | 34 s |
| Zstandard | 3 s |

image

Not only Zstandard has achieved the best compression ratio, but it has performed the compression in a fraction of the time required by the other two algorithms.

ziotom78:
There is a problem here, because the cache gets saved and restored as desired, but it is never accessed when running the `Mbs` module.

After a few hours of debugging, I discovered that this happens because the environment variable `PYSM_LOCAL_DATA` (used to implement caching) is being overwritten in [`mbs/mbs.py`](https://github.com/litebird/litebird_sim/blob/cache_pysm3/litebird_sim/mbs/mbs.py#L568).

@NicolettaK, if I understand correctly, you are using this feature so that we can pass our own CMB realizations to PySM3, is this correct?

ziotom78:
@NicolettaK , may you please have a look at [mbs.py, line 568](https://github.com/litebird/litebird_sim/blob/cache_pysm3/litebird_sim/mbs/mbs.py#L568)? I would like to remove the line where the code changes the value of the environment variable `PYSM_LOCAL_DATA`, but I am unsure how to do so in a way that lets `Mbs` keep working as expected.

ziotom78:
Testing the use of `Path.absolute()` after a suggestion in [PySM issue #102](https://github.com/galsci/pysm/issues/102).

PR #149, opened 2021-12-22
Switched non-solver parameters with solver parameters at line 275 of …

sgiardie:
The parameters filled at line 275 of hwp_sys are the solver parameters, so they are the ones with a final s.

Fixed inconsistency when referring to self.sim.parameter_file.

### DIFF
--- a/litebird_sim/hwp_sys/hwp_sys.py
+++ b/litebird_sim/hwp_sys/hwp_sys.py
@@ -272,7 +272,7 @@ class HwpSys:
 
         if self.correct_in_solver:
             if self.integrate_in_band_solver:
-                self.h1, self.h2, self.beta, self.z1, self.z2 = np.loadtxt(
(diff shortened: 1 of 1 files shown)


ziotom78:
Hi @sgiardie , I noticed that the build is failing because of a trivial formatting issue. I suggest you use `pre-commit` as described in our [docs](https://litebird-sim.readthedocs.io/en/master/installation.html#run-code-validators): it will validate the code automatically every time you run `git commit`. It is so handy!

sgiardie:
Hi @ziotom78, thanks for your suggestion! I installed pre-commit and just run `pre-commit run --all-files`, which reformatted `litebird_sim/hwp_sys/hwp_sys.py`, then commited again.

paganol (approved):

paganol:
Hi @sgiardie, thanks for tracking this issue. I think it can be merged.

Issue #134, opened 2021-09-21
Split noise generation in two functions


ziotom78:
Noise creation should be done through two functions:

1. One function should add the noise to a generic NumPy array (already allocated)
2. Another function should wrap the first and accept a `Observation` object as parameter.

PR #151, opened 2022-01-13
Uniform the API of the noise generation module

ziotom78:
This PR introduces some breaking changes to the noise module. Specifically:

- There are *two* functions to add noise: `add_noise` has been renamed to `add_noise_to_observations` (similar to `add_dipole_to_observations`), and a new function `add_noise` now works on plain NumPy arrays
- Thanks to Numba, the 1/f noise generation is now ~10% faster and uses ~25% less memory
- The Simulation object now creates a new number generator (based on PCG64) for each MPI process
- New tests have been added for the noise generator (covering MPI as well)
- The functions `add_white_noise` and `add_one_over_f_noise` are now correctly exported


ziotom78:
@paganol , apart from some speed/memory optimizations, this should fix #134. Does it look reasonable?

Unfortunately, to match the API used by other modules, I had to introduce a few breaking changes: `add_noise` is now `add_noise_to_observations`, and the parameter `noisetype` has been renamed `noise_type`.

paganol:
Hi @ziotom78. It looks great to me. I made few corrections in the documentation. The only substantial change I made is that I put to zero the frequency zero in the 1/f model. https://github.com/litebird/litebird_sim/blob/861d4c311371561297ded0cc7fa41dcc86fcfdb8/litebird_sim/noise.py#L47
If this is ok for you, we can merge it

PR #152, opened 2022-01-14
Perform a critical update of NumPy and update Numba and Rich

ziotom78:
Dependabot discovered that we depend on a NumPy version (1.20.3) that has a critical bug:

https://github.com/litebird/litebird_sim/security/dependabot/poetry.lock/numpy/open/update-logs/172230966

﻿In this PR I have updated NumPy to [1.21](https://numpy.org/devdocs/release/1.21.0-notes.html). This required to upgrade Numba from 0.54 to 0.55.

I updated Rich from version 6.2 (old!) to version 11.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # HEAD
 
+-   Upgrade NumPy from 1.20 to 1.21, Numba from 0.54 to 0.55, Rich from 6.2 to 11.0 [#152](https://github.com/litebird/litebird_sim/pull/152)
+
 # Version 0.4.0
 
 - **Breaking change** Drop support for Python 3.6, enable Python 3.9 [#136](https://github.com/litebird/litebird_sim/pull/136)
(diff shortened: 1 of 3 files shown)


mreineck:
Hi @ziotom78 , the link above gives a 404 for me ... do you have another link describing the numpy bug perhaps? (I'd like to check whether I have to update my own package dependencies ...)

ziotom78:
Mmm… Unfortunately the link disappeared, I don't know what happened but the page with Dependabot «closed» issues is empty:

https://github.com/litebird/litebird_sim/security/dependabot

When I accepted the warning, I was assuming that it was related to the big yellow warning at the top of the [1.21](https://numpy.org/devdocs/release/1.21.0-notes.html) release page:

> There are unresolved problems compiling NumPy 1.20.0 with gcc-11.1.
>
> - Optimization level -O3 results in many incorrect warnings when running the tests.
>
> - On some hardware NumPY will hang in an infinite loop.

mreineck:
Ah, I see, thanks!

OK, that only seems to be an issue for people who compile their own `numpy` and use a specific compiler. Probably not a big reason to worry for us, but updating the requirement to 1.21 certainly won't hurt!

PR #160, opened 2022-04-04
Include bandpass class

giuspugl:
As we decided last week, i open this PR to include a new class in the instrument configuration encoding bandpass. 
the class  encodes several methods  e.g. : 
 - to build top-hat bandpasses 
 - to build chebyshev bandpasses
 - to apodize the edges of a tophat band with exp and cosine profiles 
 - to  generate bandpasses from a model w/ bootstrap resampling .
 
Can further improve the documentation, but feel free to start the review!


litebird_sim/bandpasses.py
```diff
@@ -0,0 +1,275 @@
...
+    This  class   encodes the basic properties of a frequency band.
+    It can be initialized in three ways:
+    - Through the default constructor, will assume top-hat bandpasses
+    provided the band centroid and the width are issued.
+    - Through the class method :meth:`.from_imo`, which reads the
+      detector bandpass from the LiteBIRD Instrument Model (see
+      the :class:`.Imo` class).
+
+    Args:
+        - nu_center (float):  center frequency in GHz
```
ziotom78, line 25: We use the convention that measure units are appended to the arguments/variable members rather than just specified in the code. In this way, the following code is clear even without resorting to the documentation:

```python
band = BandPassInfo(nu_center_ghz=30.0)  # Instead of nu_center=30.0
```

litebird_sim/bandpasses.py
```diff
@@ -0,0 +1,275 @@
...
+        - bandwidth (float):  width of the band (default=0.2)
+        - range_offset (int): increase the freq. range by an offset (default +- 10 from the edges)
```
ziotom78, line 27: The parameter `range_offset` is not clear: is it an integer because it specifies the number of samples to add to `nsamp`? (It does not seem so, but I confess I do not completely understand the logic.)

litebird_sim/bandpasses.py
```diff
@@ -0,0 +1,275 @@
...
+        if self.bandtype == "top-hat":
+            self.get_top_hat_bandpass(normalize=self.normalize)
+        elif self.bandtype == "top-hat-cosine":
+            self.get_top_hat_bandpass(apodization="cosine", normalize=self.normalize)
+        elif self.bandtype == "top-hat-exp":
+            self.get_top_hat_bandpass(apodization="exp", normalize=self.normalize)
+        elif self.bandtype == "cheby":
+            self.get_chebyshev_bandpass(normalize=self.normalize)
+        else:
+            print(f"{self.bandtype} profile not implemented. Assume top-hat")
```
ziotom78, line 68: Better use `logging.warning` instead of `print`: in this way, the user can tune the way the message is reported.

litebird_sim/bandpasses.py
```diff
@@ -0,0 +1,275 @@
...
+    def get_edges(self):
+        """
+        get the edges of the tophat band
+        """
+        return (
+            self.bandcenter_ghz * (1 - self.bandwidth_ghz / 2),
+            self.bandcenter_ghz * (1 + self.bandwidth_ghz / 2),
+        )
+
+    def get_top_hat_bandpass(self, normalize=False, apodization=None):
```
ziotom78, line 80: Is this function supposed to be called outside of the class? If not, better to rename it to `_get_top_hat_bandpass`, so that users are not tempted to use it.

litebird_sim/bandpasses.py
ziotom78: I see that there are other `print` in the code below. The same applies to them, of course.

litebird_sim/bandpasses.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+        # interpolate the inverse CDF
+        self.Sampler = sp.interpolate.interp1d(
+            Pnu,
+            self.freqs[:-1] + np.diff(self.freqs),
+            bounds_error=False,
+            fill_value="extrapolate",
+        )
+
+    def bandpass_resampling(self, bstrap_size=1000, nresample=54, model=None):
```
ziotom78, line 238: The docstring should mention what `model` is supposed to be. (Better also to add a type hint, e.g., `model: Optional[str] = None`.)

ziotom78 (commented):
Hi Giuseppe, nice work! The only things that are missing are a few tests and some paragraphs in the user's manual. I can help you with the latter, if you want.

litebird_sim/bandpasses.py
giuspugl: ok i agree!

litebird_sim/bandpasses.py
giuspugl: Yeah you are right it is a bit awkward parameter, 
`range offset` essentially defines a frequency range  it must be slightly  wider  than the bandwidth so that we are able include the apodization profile .

litebird_sim/bandpasses.py
giuspugl: ok will do that!

litebird_sim/bandpasses.py
giuspugl: It might, but for time being i will make it private yes..

litebird_sim/detectors.py
```diff
@@ -74,6 +75,11 @@ class DetectorInfo:
         - bandwidth_ghz (float): The bandwidth of the detector, in
              GHz. The default is 0.0
 
+        - band_freqs_ghz (float): band sampled frequencies, in GHz.
+             The default is None
+
+        - band_weights (float): band profile. The default is None
```
ziotom78, line 81: The docstring says that these values are `(float)`, but I think they are *arrays* of floats, correct?

giuspugl: see  `5141dc060`

litebird_sim/bandpasses.py
```diff
@@ -0,0 +1,275 @@
...
+
+        # we extend the wings of the top-hat bandpass  with 10 samples
+        # before  and after the edges
+        bandrange = self.f0 - self.range_offset, self.f1 + self.range_offset
+        self.freqs_ghz = np.linspace(bandrange[0], bandrange[1], self.nsamp)
+        self.isnormalized = False
+        if self.bandtype == "top-hat":
+            self.get_top_hat_bandpass(normalize=self.normalize)
+        elif self.bandtype == "top-hat-cosine":
+            self.get_top_hat_bandpass(apodization="cosine", normalize=self.normalize)
```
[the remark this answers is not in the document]
giuspugl: see https://github.com/litebird/litebird_sim/pull/160/commits/495bf106e0bf864c7111cea1c59b28027fb90619

ziotom78:
Hi Giuseppe,

  Thanks a lot! But there are a few things that still need to be fixed:

-   I don't understand why, but GitHub skipped almost all the CI tests: it only ran the Readthedocs build. (And your code is not properly formatted with `black`, nor it passes `flake8` checks, so I was expecting those builds to fail.) Could it be that the repository you forked disabled them?
-   We should add a manual page for bandshapes.

I have corrected all the errors (GitHub builds, formatting, `flake8` warnings) and have added the skeleton of a page in the manual, but `git` says that I do not have the right to push modifications to this PR, probably because your fork was based on Tommaso's fork.

Perhaps the quickest way to fix this is to add me as a contributor of your repository, so that I can push my fixes here.

paganol:
Hi @giuspugl and @ziotom78, we should add also an interface with Mbs, right?

ziotom78:
No, wait, this PR was not ready, the documentation is still missing!

Issue #137, opened 2021-09-24
Add e2g rotation in scan_map


paganol:
This issue tracks one feature not implemented in #131, i.e. ecplitic to galactic rotation in [scan_map.py](https://github.com/litebird/litebird_sim/blob/master/litebird_sim/scan_map.py), so to have the option of inputting a map in galactic coordinate to `scan_map_in_observations`.

PR #164, opened 2022-04-26
Galactic coordinates in scan map

paganol:
This PR addresses issue #137.  I included the pointing rotation in the `scan_map` function and, I added some new functions in coordinates.py, they are mostly adapted from [rotator.py](https://github.com/healpy/healpy/blob/main/healpy/rotator.py) of healpy.

### DIFF
--- a/litebird_sim/coordinates.py
+++ b/litebird_sim/coordinates.py
@@ -43,10 +43,11 @@ def vec2ang(vx, vy, vz):
 
 
 def rotate_coordinates_e2g(pointings):
+    curr_pointings = np.empty_like(pointings)
     vec = np.tensordot(e2g, ang2vec(pointings[:, 0], pointings[:, 1]), axes=(1, 0))
     north_pole = np.tensordot(e2g, [0.0, 0.0, 1.0], axes=(1, 0))
     sinalpha = north_pole[0] * vec[1] - north_pole[1] * vec[0]
     cosalpha = north_pole[2] - vec[2] * np.dot(north_pole, vec)
-    pointings[:, 0], pointings[:, 1] = vec2ang(vec[0], vec[1], vec[2])
-    pointings[:, 2] += np.arctan2(sinalpha, cosalpha)
-    return
+    curr_pointings[:, 0], curr_pointings[:, 1] = vec2ang(vec[0], vec[1], vec[2])
+    curr_pointings[:, 2] = pointings[:, 2] + np.arctan2(sinalpha, cosalpha)
+    return curr_pointings
--- a/litebird_sim/coordinates.py
+++ b/litebird_sim/coordinates.py
@@ -87,14 +87,14 @@ def rotate_coordinates_e2g(pointings_elc):
     Parameters
     ----------
     pointings_elc : array
-      ``(N × 3)`` array containing the colatitude, the longitude and the polarization angle
-      (in radians) in ecliptic coordinates
+      ``(N × 3)`` array containing the colatitude, the longitude and the polarization
+      angle (in radians) in ecliptic coordinates
 
     Returns
     -------
     pointings_gal : array
-      ``(N × 3)`` array containing the colatitude, the longitude and the polarization angle
-      (in radians) in galactic coordinates
+      ``(N × 3)`` array containing the colatitude, the longitude and the polarization
+      angle (in radians) in galactic coordinates
 
     See Also
     --------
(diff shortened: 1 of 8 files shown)


litebird_sim/coordinates.py
```diff
@@ -1,9 +1,115 @@
...
+    See Also
+    --------
+    https://github.com/healpy/healpy/blob/main/healpy/rotator.py#L657
+    """
+    ct, st, cp, sp = np.cos(theta), np.sin(theta), np.cos(phi), np.sin(phi)
+    vec = np.empty((3, ct.size), np.float64)
+    vec[0, :] = st * cp
+    vec[1, :] = st * sp
+    vec[2, :] = ct
+    return vec.squeeze()
```
ziotom78, line 52: This looks equivalent to [`healpy.ang2vec`](https://healpy.readthedocs.io/en/latest/generated/healpy.pixelfunc.ang2vec.html) and [`ducc0,healpix,ang2vec`](https://gitlab.mpcdf.mpg.de/mtr/ducc/-/blob/ducc0/python/test/test_healpix.py#L102-L106) (but the latter has a slightly different syntax, as it requires a N×2 matrix instead of two vectors). Why was it redefined here?

litebird_sim/coordinates.py
```diff
@@ -1,9 +1,115 @@
...
 from astropy.coordinates import BarycentricMeanEcliptic
+from astropy.coordinates import SkyCoord
 
 """The coordinate system used by the framework"""
 DEFAULT_COORDINATE_SYSTEM = BarycentricMeanEcliptic()
 
 """The time scale used by the framework"""
 DEFAULT_TIME_SCALE = "tdb"
+
+e2g = (
```
ziotom78, line 13: As it is a constant, it should use uppercase letters. Moreover, given that it might be used elsewhere, better to use a more descriptive name, like `ECL_TO_GAL_ROT_MATRIX`

litebird_sim/coordinates.py
paganol: It is. I just wanted to have a self-consistent module, but if you prefer I can use the healpy function. I must admit that I didn't check ducc0. Btw, in principle also for `rotate_coordinates_e2g` we could use the functions of healpy.rotator. Again if you think it is better to go for the healpy functions, I can rework the module.

litebird_sim/coordinates.py
```diff
@@ -1,9 +1,121 @@
...
+      The angles in radiants in an array of shape (2, N)
+
+    See Also
+    --------
+    https://github.com/healpy/healpy/blob/main/healpy/rotator.py#L610
+    """
+
+    r = np.sqrt(vx ** 2 + vy ** 2 + vz ** 2)
+    ang = np.empty((2, r.size))
+    ang[0, :] = np.arccos(vz / r)
```
mreineck, line 83: I suggest to use something like `ang[0] = np.arctan2(np.sqrt(vx**2 + vy**2),vz)`, which is more accurate near the poles and does not break in presence of numerical noise. See for example https://gitlab.mpcdf.mpg.de/mtr/ducc/-/blob/ducc0/src/ducc0/math/pointing.cc#L46

mreineck: My personal rule of thumb is to _never_ use `arcsin` or `arccos` anywhere, they are causing too much trouble because of their lmited definition range and their infinite derivatives near the borders. There is always a better solution using `arctan2`...

litebird_sim/coordinates.py
mreineck: `ducc0` is using the diffeent syntax to minimize the calling overhead as much as possible. It' true that this is more restrictive than the `healpy` one, and it's one of the main reasons why `healpy` isn't using `ducc0` internally as a back-end yet.

ziotom78: In fact, yesterday I did some tests and found that `ducc0.healpix.ang2vec` is ~10% faster than `healpy.ang2vec`.

ziotom78: @paganol  for me it's fine to keep our own implementation in the code. I would like to abandon healpy some time in the future in favor of ducc, so the fewer dependencies we have, the better.

sgiardie (approved):

ziotom78 (changes requested):
It looks great to me! The last thing that is still missing is to export the `ECL_TO_GAL_ROT_MATRIX` and `CoordinateSystem` symbols in `litebird_sim/init.py`, under the section for file `coordinates.py`, then it's ready to be merged.

paganol:
> We'll probably need to release a new version of the framework at that point.

Agree. I would just wait for the comparison that @giuspugl and @sgiardie are doing. If everything works fine we can release a new version.

Issue #141, opened 2021-10-27
Standardize the "pointings", "psi", and "pixidx" fields


ziotom78:
More and more modules in the framework are assuming that each observation has pointing information stored in these fields:

-   `pointings` (a D×N×3 matrix, with D number of detectors, N number of samples and the last axis sampling the three variables θ, φ, and ψ)
-   `pixidx` (the pixel index)

As pointings are trivially generated through a call to `litebird_sim.scanning.get_pointings`, we should provide a small function that accepts a `Observation` as input, calls `get_pointings` and stores the result in these variables.

PR #168, opened 2022-05-11
Add hwp_rad_sec keyword to get_pointings

### LINKED ISSUE #155: HWP polarization angle and get_pointings
We should include an option in get_pointings which adds twice the angle of the hwp to the polarization angle of the detectors. In this way, for an ideal case, the pointing array has already the "right" polarization angle and it's much easier to use in other modules. What do you think?

ziotom78:
This should fix #155. I use the same trick described in the issue to compute the absolute time of the HWP.

One thing that is missing is a mechanism to let non-idealities in the HWP spinning rate. Perhaps a good option might be to implement a HWP object in the code and change the signature of `get_pointings` so that it gets this object instead of the plain parameter `hwp_rad_per_sec`:

```python
hwp = IdealRotatingHWP(hwp_rad_sec=0.1521)
pointings = lbs.get_pointings(..., hwp=hwp)
```

@paganol, what do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
     -   Each `Simulation` object creates random number generators (field `Simulation.random`), in a way that is safe even for MPI applications
 
+-   Use a more robust algorithm to compute pointings [#175](https://github.com/litebird/litebird_sim/pull/175)
+
 -   Improve the documentation for the destriper [#172](https://github.com/litebird/litebird_sim/pull/172)
 
 -   Make `scan_map_in_observations` and `add_dipole_to_observations` accept list of pointing matrices [#171](https://github.com/litebird/litebird_sim/pull/171)
--- a/litebird_sim/scanning.py
+++ b/litebird_sim/scanning.py
@@ -170,7 +170,7 @@ def compute_pointing_and_polangle(result, quaternion):
     # vectors before the final result. First, we use it to compute the
     # (x, y, z) pointing direction
     rotate_z_vector(result, vx, vy, vz, w)
-    theta_pointing = np.arccos(result[2])
+    theta_pointing = np.arctan2(np.sqrt(result[0] ** 2 + result[1] ** 2), result[2])
     phi_pointing = np.arctan2(result[1], result[0])
 
     # Now reuse "result" to compute the polarization direction


paganol:
Hi @ziotom78, it looks great to me. I opened last week a branch [here](https://github.com/litebird/litebird_sim/tree/hwp_in_get_pointings) with a possible solution, but yours is much better. The only other thing I would change is scan_map.py making `hwp_radpsec` optional, like [here](https://github.com/litebird/litebird_sim/blob/hwp_in_get_pointings/litebird_sim/scan_map.py). If you like it I can move it in this PR.

ziotom78:
Oh sorry @paganol , as usual I forgot about your other PR!

I have changed the implementation so that now you can avoid specifying a HWP. there is a `HWP` abstract base class that represents the half-wave plate. For the moment, only `IdealHWP` is provided: it does the very same calculation as in your code, @paganol. The way `get_pointings` works now is shown with this example:

```python
from lbs import get_pointings, IdealHWP

pointings = get_pointings(…, hwp=IdealHWP(ang_speed_rad_sec=1.0))
```

This implementation is ready to be extended with other features:

1.  We can add a static method `IdealHWP.from_imo` that reads the HWP parameters from the IMO (we surely want this!)
2.  We can store other fields in `HWP`, like Müller matrices
3.  We can implement other derived class, like `NonIdealHWP`, .e.g., to include non-constant angular speeds, etc.

What's missing is some documentation, but first I would like to know if this implementation looks ok.

paganol:
Do we want to remove the support for the hwp rotation from scan_map, or keep it as optional?

ziotom78:
> Do we want to remove the support for the hwp rotation from scan_map, or keep it as optional?

I would like to remove the parameter from `scan_map`: it would look confusing to users, it does not easily enable the simulation of systematics in the angular rotation, and it does not fit nicely with the IMO api.

If you agree, I can remove it and fix the tests.

paganol:
Sure, my idea was not to keep hwp_rad_sec as it is, but pass `HWP` as optional
Something like this 
```
def scan_map_in_observations(
    obs: Union[Observation, List[Observation]],
    pointings,
    maps: List,
    hwp: Optional[HWP] = None,
    input_map_in_galactic: bool = True,
    fill_psi_and_pixind_in_obs: bool = False,
):
```

with this in `scan_map`

```
    if hwp:
        hwp.add_hwp_angle(
            pointings,
            start_time_s,
            delta_time_s,
        )
```

But I understand if you prefer to remove this possibility. I don't see for now a situation where keeping this option has a value. My point was, simply, let's keep it, maybe one day it will be useful.

paganol:
bwt, do we want to fill obs.pixidx and obs.psi inside get_pointings? It seems the right moment and the right PR, this sould fix issue #141

ziotom78:
> Do we want to remove the support for the hwp rotation from scan_map, or keep it as optional?

I would like to remove the parameter from `scan_map`: it would look confusing to users, it does not easily enable the simulation of systematics in the angular rotation, and it does not fit nicely with the IMO api.

If you agree, I can remove it and fix the tests.

> bwt, do we want to fill obs.pixidx and obs.psi inside get_pointings? It seems the right moment and the right PR, this sould fix issue #141

Yes, that's right, I have changed the implementation so that `pointings` and `psi` are saved in a observation. Not sure about `pixidx`, though:

1. At the moment, `get_pointings` only computes the *angles* and not the pixels. Making it convert directions into pixels means that one should provide a value for `NSIDE` to `get_pointings`, which is not always what one wants (e.g., simulation of planetary visibilities)
2. If we save pixel indices in a `Observation`, we should save the associated `NSIDE` and the ordering scheme (RING/NESTED), otherwise the information is useless.
3. The value of NSIDE and the ordering scheme should be saved in HDF5 files as well, but at the moment this is still missing.

Thoughts?

ziotom78:
> Ok, I agree. I'm trying to modify scan_map accordingly, what should it take? Should it assume that the pointing and polarization angles are stored in obs, or should we pass obs and pointings separated like it is now?
> 
> ```
> def scan_map_in_observations(
>     obs: Union[Observation, List[Observation]],
>     pointings: Union[np.ndarray, List[np.ndarray]],
>     .....
> ```
> 
> On the same line, now `get_pointings` with `store_pointings_in_obs=True` returns the pointing_buffer, effectively storing the same information twice. Should we instead return an empty buffer if `store_pointings_in_obs=True`?

I think that the best option would be to leave the parameters optional, and if they are None they are taken from the observation (much like quaternions in `get_pointings`). There is no memory wasted in returning the pointing information from `get_pointings`, as NumPy arrays are reference counted, and thus the same array is stored in `Observation` and returned, i.e., there is no double allocation.

ziotom78:
@paganol I fixed a few problems in the PR, but there is still an issue with ducc0. I have asked Martin privately for a solution.

PR #169, opened 2022-05-11
High pass filter for the noise

paganol:
The noise generator now accepts the fmin parameter, as specified in the IMO. Also, `add_noise_to_observations` accepts a single obs or a list of obs similarly to `add_dipole_to_observations` and `scan_map_in_observations`.


ziotom78:
Hi @paganol , this looks fine to me, please merge it if you feel it's ready.

PR #176, opened 2022-06-06
Pixel index computation in make_bin_map

paganol:
This PR makes mapping compatible with the new version of scan_map


ziotom78:
It looks good! Feel free to merge when you want.

PR #178, opened 2022-06-07
Typos, test and some documentation

paganol:
This PR corrects a typo in `make_bin_map`, improves the test `test_mapping.py` and adds some documentation

### DIFF
--- a/docs/source/mapmaking.rst
+++ b/docs/source/mapmaking.rst
@@ -88,9 +88,13 @@ simple binned map with the function :func:`.make_bin_map`. This function
 takes: a single (or a list) of :class:`.Observations`, the Healpix
 resolution of the output map (``nside``) and produces a coadded map.
 It assumes white noise and each detector gets weighted by 
-:math:`1 / NET^2`. Optionally, if the parameter do_covariance is True,
-it can output the covariance matrix in an array of shape 
-`(12 * nside * nside, 3, 3)`. This is how should be called::
+:math:`1 / NET^2`. If the pointing information is not provided in the 
+observation, it can be passed through the optional argument `pointings`, 
+with a syntax similar to :func:`.scan_map_in_observations`.
+The output map is in Galactic coordinates, unless the optional prameter
+`output_map_in_galactic` is set to False. If the parameter `do_covariance` 
+is True, it can output the covariance matrix in an array of shape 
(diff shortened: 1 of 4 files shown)


litebird_sim/destriper/__init__.py
```diff
@@ -216,11 +217,15 @@ def __init__(
         )
 
     def local_intervals(self, _):
+        start_time = (
```
sgiardie, line 220: Hi @ziotom78, I added this small piece here to handle obs.start_time passed as an astropy.time.Time object. In this case, `obs.start_time` is not a number and we have to call `obs.start_time.cxcsec`

paganol:
Hi @ziotom78, we changed a bit the destriper, now it retrives the pointing from the observations (or pointings can be passed as optional argument). The previous version failed with a rotating hwp. If you like the last changes we can merge and release 0.5.0. I think @sgiardie has also an example notebook ready.

sgiardie:
Hi @ziotom78 and @paganol, I did some small changes in `destriper/__init__.py` to solve some errors due to the shape of `sim.observations` and `pointings`. I also added the notebook I have been working on. Now the binned map from `make_bin_map` and the one coming from MADAM seem to agree quite well, instead the destriped map is a bit weird. I plotted that in the notebook, if you can give a look at it you can tell me what is wrong and what should I change

sgiardie:
I reverted the changes in `destriper/__init__.py`, since the errors I was having depended on not passing `pointings` as a list of pointings corresponding to each `sim.observations`. I should have done 
```
pointings=[]
for obs in sim.observations:
    pointings.append(lbs.get_pointings(obs,
                      spin2ecliptic_quats=sim.spin2ecliptic_quats,
                      detector_quats = None,
                      bore2spin_quat=instr.bore2spin_quat,
                      hwp=lbs.hwp.IdealHWP(hwp_radpsec),
                     ))
```
as I am doing now. 
The notebook is corrected accordingly, but of course the issue in the destriped maps is still there

sgiardie:
Hi @ziotom78, I removed the destriper examples from the notebook. You may want to reformat it again, though. I was not able to do that

ziotom78:
Hi Serena, I checked and the code looks ok, black does not need to change it. You can merge it, thanks!

PR #182, opened 2022-07-07
Fix the way the mapmaker handles signals

ziotom78:
This PR modifies `_Toast2FakeCache` so that it performs deep copies even when `Signal.copy()` is called.


paganol:
Hi @ziotom78 @giuspugl @sgiardie, currently `baseline_lenght` is interpreted as the number of consecutive samples in a baseline. Digging in toast, it seems that it is actually the length in seconds. @giuspugl, do you confirm this?

giuspugl:
> Hi @ziotom78 @giuspugl @sgiardie, currently `baseline_lenght` is interpreted as the number of consecutive samples in a baseline. Digging in toast, it seems that it is actually the length in seconds. @giuspugl, do you confirm this?

I confirm that it should be the length in seconds !

ziotom78:
@paganol , @sgiardie , what do you think about the breaking change? I would be in favor of implementing it and releasing 0.6.0, as in this way the simulation production will have a self-explaining code.

sgiardie:
> @paganol , @sgiardie , what do you think about the breaking change? I would be in favor of implementing it and releasing 0.6.0, as in this way the simulation production will have a self-explaining code.

Hi @ziotom78 , sorry for the late reply. I think it is a reasonable change, and also not so impacting on the production scripts (it would require just to rename that variable, I guess)

ziotom78:
> (it would require just to rename that variable, I guess)

Yes, that's right, it's something you can do automatically if you use an IDE like PyCharm or Spyder.

I'll draft release v0.6.0 then, instead of v0.5.1, as this is a breaking change.

ziotom78:
> I'll draft release v0.6.0 then, instead of v0.5.1, as this is a breaking change.

Wait, I just realized that we can keep `baseline_length` as well but mark it as «deprecated». In this way we can release 0.5.1 as planned.

ziotom78:
> Wait, I just realized that we can keep `baseline_length` as well but mark it as «deprecated». In this way we can release 0.5.1 as planned.

Mmm, no, deprecating a field of a dataclass is not trivial and I fear to complicate the code too much for little gain.

Sorry for the noise.

PR #186, opened 2022-08-02
Add support to call Madam mapmaker (externally)

ziotom78:
This PR implements a new function, `save_simulation_for_madam`, which saves the TODs and pointings for a set of detectors to FITS files, alongside with a `.par` and a `.sim` file that can be fed to Madam's mapmaker.

- [X] Create `save_simulation_for_madam`
- [X] Move `DestriperParameters` and `DestriperResults` out of the TOAST2 wrapper, so that the can be re-used again by other map-makers
- [X] Add tests
- [X] Add some documentation


ziotom78:
Since there have been no negative comments, I am merging this to `master`.

PR #191, opened 2022-09-01
Allow nnz=1 in the destriper

ziotom78:
This small PR lets the destriper run if one specifies `nnz=1` (i.e., only solve for intensity I).


ziotom78:
Now the part modifying the dependencies has been incorporated in #192, so that the history for this PR will be cleaner when merged.

PR #192, opened 2022-09-07
Fix dependencies

ziotom78:
- [X] Remove old dependency (was used to support Python 3.6)
- [X] Update poetry.lock
- [ ] Fix problem with sphinxcontrib.asciinema


ziotom78:
This PR incorporates those changes made in #191 to make the CI builds pass, to make the repository history clearer.

PR #193, opened 2022-09-07
Make the CI builds test the latest stable Python release (3.10)

ziotom78:


ziotom78:
Unfortunately, it seems that AstroPy 4 does not support Python 3.10, but we cannot upgrade to AstroPy 5 because it drops support for Python 3.7 (which we do still support).

Issue #197, opened 2022-09-19
Documentation for hwp.py


paganol:
We should write a minimal tutorial on the usage of get_pointings in case of HWP and some documentation for the class IdealHWP.

paganol:
Vaguely related, should detector_quats in get_pointings be an optional argument with None as default? So that, by default, the code takes the quaternions stored in the observation. Again, should we have a wrapper function that takes a list of observations and produces a list of pointings? Instead of being forced to do something like this
```
pointings=[]
for obs in sim.observations:
    pointings.append(lbs.get_pointings(obs,
                      spin2ecliptic_quats=sim.spin2ecliptic_quats,
                      detector_quats = None,
                      bore2spin_quat=instr.bore2spin_quat,
                      hwp=lbs.hwp.IdealHWP(hwp_radpsec),
                     ))
```

PR #198, opened 2022-09-20
hwp doc and change in get_pointing

paganol:
This PR addresses the issues described in #197. 
I made the following changes:
- In `get_pointings`, the parameter `detector_quats` became optional. Now, not passing this parameter, the function takes detector quaternions from the observation passed.
- A new function was created, called `get_pointings_for_observations`, which wraps `get_pointings`. Now you can pass a list of observations to this function and a list of pointings is returned. This function provides a nice shortcut with a simplified interface with fewer options. Of course, for a more refined simulation, `get_pointings` must be used. The logic is the same we used in `add_noise_to_observations`, `scan_map_in_observations`, etc... 
- Relevant tests were also fixed and they all pass on my laptop. On GitHub, they fail, but it seems to be a problem not related to my commits.
- The documentation is fixed accordingly. I also added a small extra entry in the section [Scanning strategy](https://litebird-sim.readthedocs.io/en/master/scanning.html) for describing how to use the `HWP` class with `get_pointings`.


ziotom78:
@paganol , sorry for the mess, I mistakenly merged #160 here and had to revert it. Now the CI builds pass, this PR can be merged.

PR #203, opened 2022-10-11
tutorial notebook on bpasses

giuspugl:
including a notebook on the usage of Bandpasses


ziotom78:
Yes, right, I remember that I converted the notebook to a tutorial with some theory and a few equations for the mathematical model:

https://litebird-sim.readthedocs.io/en/latest/bandpasses.html

If you think it would be useful to have a notebook as well, it's fine to merge it, but in this case you should remove references to your own filesystem and make it runnable on other systems as well. (There is a reference to a TOML file located at `/Users/peppe/work/satellite_sims/hardware/MFT_subset.toml` in cell 3.)

giuspugl:
with the previous commits we address the comments on this PR and also some slight change in the warning message to make it more understandable and a minor fix in the extrapolation values. 
we 're ready to merge.

PR #208, opened 2022-11-16
Add flag for coordinate system choice of madam output maps

marcobortolami:
Now, the `save_simulation_for_madam` function does not allow to choose, in the `madam.par` parameter file, the coordinate system of the maps produced by the Madam mapmaker. This PR adds the possibility to tell Madam to produce maps in Ecliptic or Galactic coordinates.

### DIFF
--- a/litebird_sim/madam.py
+++ b/litebird_sim/madam.py
@@ -10,6 +10,7 @@ import jinja2
 
 import litebird_sim
 from . import DetectorInfo
+from .coordinates import CoordinateSystem
 from .mapping import DestriperParameters
 from .mpi import MPI_COMM_WORLD
 from .observations import Observation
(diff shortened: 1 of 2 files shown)


marcobortolami:
@nraffuzz and I have already performed a test setting `coordinate_system=lbs.coordinates.CoordinateSystem.Galactic` when creating an instance of `lbs.DestriperParameters`. The `madam.par` parameter file correctly shows `to_galactic = T` and, by running madam, the output maps are in galactic coordinates. If you agree with our changes, we could merge this PR.

ziotom78:
Thanks, I don't know how I could miss that…

Tests are not passing because you must run

    black .

on the source to reformat the new line you added to `madam.py`, and then  `git add` it again.

Issue #209, opened 2022-11-17
Issue with make_bin_map after PR205


marcobortolami:
@martamonelli, @nraffuzz and I noticed something when running `e2e_simulations.py` modified after #205.
- If we compare the spectra of output (cmb, fg, wn) and input maps with the output one produced by the binned mapmaker, the spectra DO NOT overlap, and the ratio in/out is ~9.
- If we compare the spectra of output and input maps with the output one produced by madam (using the binned version), the spectra DO overlap. So, no problem with madam.

We remembered about a similar problem: some weeks ago we were passing a list of observations and not a single observation containing the sum of the tod components in the `obs.tod` field and the problem was exactly the same. For that case it was like the simulation time was multiplied by the number of components used.
In order to test whether this is the problem, we produced the following maps with `lbs.make_bin_map`:
1) cmb
2) cmb + fg
3) cmb + fg + wn

and we get the following results:
```
1) Cl_cmb_input   / Cl_cmb_binned        ~ 1
2) Cl_coadd_input / Cl_cmb_fg_binned     ~ 4
3) Cl_coadd_input / Cl_cmb_fg_wn_binned  ~ 9
```
So, in conclusion, our guess is that, when using more than 1 component in `lbs.make_bin_map` after #205, it is like the simulation time gets multiplied by the number of components. We got correct results when we were using one single observation but summing all the components in `obs.tod`.
What do you think?

ziotom78:
Sorry, I did a stupid mistake and was too lazy to skip the implementation of a proper test, silly me!

PR #210 should fix the issue, may you please test it?

marcobortolami:
PR #210 solves the issue, now all the spectra overlap. Thanks for the quick action!

PR #210, opened 2022-11-17
Fix issue #209

ziotom78:
This should fix issue #209. The PR includes a test to verify that the solution of the map-maker works with more than one component. (I should have implemented it before!)


marcobortolami:
This PR solved issue #209.

PR #212, opened 2022-11-28
Check that non-standard TOD components are filled properly

ziotom78:
This PR is an attempt to fix the weird behaviour seen in E2E tests and reported at the link https://github.com/litebird/e2e-simulation/issues/3 . It includes a couple of tests that verify that the `component=` keyword is honored.


ziotom78:
> mmmh... I've just done a small fix to a comment using the GitHub online editor and now some GitHub tests are stalling, I don't know why. The test `Tests / tests (ubuntu-latest, mpich, 3.7) (push)` runs without problems but stalls after `test/test_scanning.py::test_simulation_two_detectors PASSED [ 82%]`

Yes, it happens sometimes… A few VMs used by GitHub get stuck, and re-running those jobs don't help.

Anyway, the tests pass without problems on my computer and on several VMs, so I do not really know why you were experiencing those strange leaks in the signals… May you please try to add some new test (similar to the ones I added) able to reproduce your problem?

PR #216, opened 2022-12-19
Temporarily disable mpich

ziotom78:
It has been a few weeks that tests using mpich take forever to compile. It's hard to investigate this, because it might likely be due to some outdated dependencies. I would prefer to disable mpich tests until we deprecate Python 3.7 and update all our dependencies; keeping mpich tests enabled might suck up all the free time to run CI builds.


ziotom78:
I'm merging this, but let's remember to revert this change once we deprecate support for Python 3.7 and review all the dependencies.

PR #221, opened 2023-02-02
New functions in simulations

paganol:
I've added some new functions that should simplify the generation of simulations. @ziotom78, what do you think?

Still few things should be done:
- [x] Add a similar function also for the noise
- [x] Add documentation
- [x] Fix the automatic report
- [x] Update the notebook, with the new syntax. We can probably create a new one, and keep track of the more flexible functions
- [x] And, probably, find better names ;-)


paganol:
btw, I've also added in the dictionary returned by Mbs a new field with the info about the coordinates. This might be useful in the future. I'm using it in scan_map for now.

litebird_sim/scan_map.py
```diff
@@ -90,7 +90,7 @@ def scan_map_in_observations(
     obs: Union[Observation, List[Observation]],
     maps: Dict[str, np.ndarray],
     pointings: Union[np.ndarray, List[np.ndarray], None] = None,
-    input_map_in_galactic: bool = True,
+    input_map_in_galactic: Union[bool, None] = True,
```
ziotom78, line 93: What's the meaning of `None` here? Is `input_map_in_galactic` becoming a three-state variable? (`True`/`False`/`None`.) Perhaps the meaning of `None` should be documented in the docstring.

ziotom78:
@paganol , that's very nice! When you think it's ready, feel free to merge it.

PR #225, opened 2023-03-02
Better handling of multiple TODs

ziotom78:
This PR implements a high-level API to allocate multiple TODs during a call to `Simulation.create_observations`: the new parameter `tods` lets the caller to specify how many TOD matrices must be allocated, their `dtype`, and a short description. This information is used when TODs are saved to HDF5 files and is restored properly when those files are read.

- Add TodDescription and the ability to allocate multiple TODs in Observations
- Support multiple TODs in `Simulation` objects
- Make a few tests faster
- Improve the API to save multiple TODs in HDF5 files
- Add tests for the multi-TOD HDF5 API
- Reformat notebook
- Fix tests and API for Madam


ziotom78:
This PR should be complete. The new API lets the user to allocate several TODs at once:

```python
    sim.create_observations(
        detectors=[det1, det2, det3],
        tods=[
            lbs.TodDescription(
                name="tod", description="TOD", dtype=np.float64,
            ),
            lbs.TodDescription(
                name="noise", description="1/f+white noise", dtype=np.float32
            ),
        ],
    )

    for cur_obs in sim.observations:
        print("Shape of 'tod': ", cur_obs.tod.shape)
        print("Shape of 'noise': ", cur_obs.noise.shape)
```

These TODs are reported by `Simulation.describe_mpi_distribution` and can be referred to when saving/loading HDF5 files.

PR #230, opened 2023-03-29
Add the PTEP IMO to the source code distribution

ziotom78:
This PR includes a “default” IMO file, containing the information that were published in the [PTEP paper](https://doi.org/10.1093/ptep/ptac150).

- [X] Add the JSON PTEP file
- [X] Add the JSON PTEP file to the list of redistributable files
- [X] Add the ability to read many (gzipped?) IMO files and merge them
- [X] Include the PTEP IMO in the distribution
- [x] Add tests
- [x] Update the CHANGELOG
- [x] Update the documentation


ziotom78:
Finally, after nearly 7 months of work, this PR is done! I have updated some tests and the documentation, so that in many places we use the PTEP IMO instead of hand-made `DetectorInfo`/`ScanningStrategy`/etc. classes. The documentation should be more readable and the tests easier to parse.

The PTEP tree structure conforms to the so-called “IMO v2.0”, which is however *not* the IMO v2.0 that was discussed at Elba but a previous version (it should have been called IMO v1.4), whose structure is similar to IMO v1.3.

Please comment, I would like to merge this PR in a few days.

Issue #241, opened 2023-05-24
TOAST interface not working with components different from obs.tod


marcobortolami:
It seems that the TOAST interface is not working if we use TOD components different from `obs.tod`, while it works using `obs.tod`. I mean that if we call the destriper like in the code below using the two problematic lines it fails, while removing the two problematic lines it works.
```
(obs_multitod,) = sim.create_observations(detectors=dets,
                                          n_blocks_det=1,
                                          n_blocks_time=size,
                                          split_list_over_processes=False)
### problem from here ###

obs_multitod.tod_cmb_fg = np.zeros_like(obs_multitod.tod)
obs_multitod.tod        = np.array([], dtype='float32') #not used, to save memory

### problem to here ###

[...]

params_madam = lbs.DestriperParameters(nside=nside,
                                       coordinate_system=lbs.coordinates.CoordinateSystem.Galactic,
                                       nnz=3,
                                       baseline_length_s=60,
                                       iter_max=100,
                                       return_hit_map=False,
                                       return_baselines_map=False,
                                       return_binned_map=True,
                                       return_destriped_map=True,
                                       return_npp=False,
                                       return_invnpp=False,
                                       return_rcond=False)

result = lbs.destripe(sim, params_madam)
```
[This line](https://github.com/litebird/litebird_sim/blob/23fff2cf6a89224f17cd504b6e9a9bdbba38effb/litebird_sim/destriper/__init__.py#LL138C41-L138C44) is causing the problem.

PR #242, opened 2023-05-24
make TOAST interface work with components different from obs.tod

marcobortolami:
This PR has been opened to solve #241.


marcobortolami:
I just added the `components` keyword to the high level functions, following what has been done for madam. We should think about the best strategy to create and use all the instances needed by TOAST

marcobortolami:
1. The usage would be easier if we just list the components that we want to use on the `destripe` function and the mapmaker handles their sum, at a cost of more memory usage
2. In the second case the user should take care of the sum, that is not a difficult task. The user could create a new TOD for the sum externally and pass just that component, or sum the interesting components in the same TOD field, saving memory, e.g.
```
obs.tod_cmb += obs.tod_fg
obs.tod_cmb += obs.tod_wn
```
3. Another solution in the middle of the two above would be checking if the list of `components` has more than 1 element. If so, a new TOD field is created for the sum. If not, that TOD is used for the mapmaking.

I would go with 2 or 3.

ziotom78:
Thanks for the comment, @marcobortolami . My preference is for option 2. Apart from pointings, TODs are the largest objects the framework keeps in memory, so asking the user to explicitly allocate a variable for the sum of the TODs might help in tracking down unexpected memory requirements or to quickly find the culprit of job failures.

@paganol , what do you think?

paganol:
I there, I agree with @ziotom78. I would go for the option 2

ziotom78:
Perfect, I implemented option #2 and updated the tests and the docstrings.

PR #245, opened 2023-06-06
Memory optimisation

paganol:
This PR reduces the memory consumption of `make_bin_map`


ziotom78:
I was able to solve the warning (and possibly make the code slightly more robust) by using `numpy.linalg.solve` instead of the combo that inverts a matrix and uses `numpy.dot` on it. (Usually solvers avoid creating the inverse matrix to reduce roundoff errors.) This has the benefit of solving the warning produced about `numpy.dot`.

I also removed a useless memory allocation, but I am not sure if this solves the original memory problem.

paganol:
Sure, that was the original implementation, and I agree that it's better. I went for inverting `ata` and then using the dot product for obtaining the covariance for free. With this last change we are not outputting the covariance matrix, ma its inverse. 

Regarding the memory, yes, nothing changed.

ziotom78:
I believe we can merge this and then release version 0.10.0. @paganol , I'll do it tomorrow in the early morning if there are no objections.

Issue #255, opened 2023-07-06
Noise default seed discussion


marcobortolami:
Hi:) I open this Issue to discuss about the seed used in the random number generator if it is not specified when, for example, the user wants to produce noise TODs.

**Description of my concern**
 In my opinion:

- if the seed **is not** specified and the user runs two times a script producing noise timelines, then the noise produced should be different.
- if the seed **is** specified and the user runs two times a script producing noise timelines, then the noise produced should be the same (two runs with the same seed of course).

However, with the current version of the code, depending on which (high or low) level function the user calls, the noise timelines produced without specifying a seed are different or equal for two runs of the same code. I will make myself more clear with the following examples.

**CASE 1: high level function called, no seed specified**
```
import litebird_sim as lbs
import numpy as np
import matplotlib.pylab as plt
from astropy import units, time

sim = lbs.Simulation(
    start_time=time.Time("2025-01-01T00:00:00"),
    duration_s=1 * units.minute.to("s"),
)

detector = lbs.DetectorInfo(
    name="foo",
    sampling_rate_hz=10.0,
    bandcenter_ghz=200.0,
    net_ukrts=50.0,
    fknee_mhz=20.0,
    fmin_hz=1e-05,
    alpha=1.0,
)

sim.create_observations(
    detectors=[detector],
)

sim.add_noise('white')

times = sim.observations[0].get_times()-sim.observations[0].start_time.cxcsec

plt.plot(times,sim.observations[0].tod[0,:])
plt.xlabel("Time [s]")
plt.ylabel("Signal [K]")
```
if you run the code above two times, the noise TODs are **the same**. This happens because [here](https://github.com/litebird/litebird_sim/blob/cacc004a3b43dfad819c4fc60435e8a71f83ca3a/litebird_sim/simulations.py#L318) `sim.random` is set to `None` in the constructor of `Simulation`, but a few lines later [here](https://github.com/litebird/litebird_sim/blob/cacc004a3b43dfad819c4fc60435e8a71f83ca3a/litebird_sim/simulations.py#L385) the function `self.init_random()` is called and two lines later you seed that the seed `12345` is used. Then, in [`sim.add_noise()`](https://github.com/litebird/litebird_sim/blob/cacc004a3b43dfad819c4fc60435e8a71f83ca3a/litebird_sim/simulations.py#L1317) we have
```
if random is None:
    random = self.random
```
and this is passed to `lbs.add_noise_to_observations()`.

[...]

PR #256, opened 2023-07-13
Fix noise seed inconsistency

marcobortolami:
This PR has been created to fix the inconsistency of the noise generation pointed out in #255.


marcobortolami:
Up to now I updated `simulations.py` and `noise.py`. The only difference with respect to what discussed in #255, I think, is that I removed the default also in [`Simulation.add_noise`](https://github.com/litebird/litebird_sim/blob/cacc004a3b43dfad819c4fc60435e8a71f83ca3a/litebird_sim/simulations.py#L1317) in addition to the functions in `noise.py`. I made this choice for a better consistency among the function/method calls that generate noise: in this way, whenever generating noise, the parameter `random=...` must always be specified. Otherwise, only for `Simulation.add_noise()` there would be the possibility not to specify a RNG. Feel free to change this:)

marcobortolami:
Now, these things remains to be done:

- [ ] Add the `random_seed` value in the automatically generated sections in the reports. In my opinion it's better to keep track of the seed in the reports;
- [ ] Add a test that checks the (non) reproducibility when passing (a `None`) an integer `random_seed`;
- [ ] Fix ALL the tests, that are failing because a breaking change has been introduced in a class used basically everywhere;
- [ ] Fix the jupyter notebooks used as tutorial;
- [ ] Fix ALL the examples in the documentation;
- [ ] Add a new page in the documentation dedicated to the noise;
- [ ] Change the slides used as tutorial for litebird_sim (e.g. for the Tokyo hands-on meeting) and related material.

I can take care of these modifications. If you want, in the meanwhile, comment the previous commits:)

marcobortolami:
I fixed most of the tests. Only the ones taking parameters from a toml file or from a dictionary are still failing, because `random_seed` is a parameter that is needed in the constructor call even if it's present in the toml file or in the dictionary.
One way around is to pass it as `None` in the constructor call, so it does not raise an error and it gets it from the toml file/dictionary later, but this may be confusing or shady.
One better solution is giving a default value to `random_seed` that is "wrong", like an empty string, and check in the constructor just before `init_random` whether `random_seed` still has that value, after collecting parameters directly passed to the constructor, parameters in a dictionary or in a toml file. If this is the case, it means that `random_seed` has not been passed and a message saying `set random_seed to None or int` should appear.
What do you think?

marcobortolami:
Since the default random number generator of `numpy` is actually PCG64 and since `SeedSequence` accepts `None`, I uniformed and simplified `sim.init_random`. I did some tests and verified that:

- Passing `None` the random numbers are different for two different ranks and for two different runs of the same script
- Passing an `int` the random numbers are different for two different ranks but the same for two different runs of the same script

litebird_sim/simulations.py
```diff
@@ -395,26 +395,32 @@ def init_random(self, random_seed):
...
 
-        # Create a list of N seeds, one per each MPI process
-        seed_seq = SeedSequence(random_seed).spawn(self.mpi_comm.size)
+            # Create a list of N seeds, one per each MPI process
+            seed_seq = SeedSequence(random_seed).spawn(self.mpi_comm.size)
 
-        # Pick the seed for this process
-        self.random = Generator(PCG64(seed_seq[self.mpi_comm.rank]))
+            # Pick the seed for this process
+            self.random = Generator(PCG64(seed_seq[self.mpi_comm.rank]))
```
ziotom78: Mmm, if I understand correctly the code, it seems that when the user sets `random_seed=None`, the random generator is no longer PCG. This means that the code is less robust when used with MPI, because it skips the call to `SeedSequence`.

I would rather keep everything as is and only change the call to `SeedSequence` so that if `random_seed=None` it passes a random number to `SeedSequence` (e.g., `np.random.default_rng().integers(0, 2**31)`). In this way, we are sure that the generator will *always* be PCG64.

ziotom78 (commented):
Nice work!

litebird_sim/simulations.py
ziotom78: Ah, great, then I guess it's ok to pass `None` to `SeedSequence`. The new code is nice, in this way it's clear we're *always* using PCG64 in any case by just reading the code.

Thanks!

marcobortolami: [Here](https://docs.scipy.org/doc/numpy-1.3.x/reference/generated/numpy.random.seed.html) I found this:

> If seed is None, then RandomState will try to read data from /dev/urandom (or the Windows analogue) if available or seed from the clock otherwise.

That I guess is something nice? Or could it be dangerous? I mean, if different CPUs look in the same place at the "same time" (maybe it's a naive sentence). I'm not sure what each CPU does... But yeah, I guess it's ok.

marcobortolami:
With the last commit I:

- Corrected 2 small bugs (I was using `random_seed` instead of `self.random_seed`);
- Went with the solution of a default value for `random_seed` that is `""` and just before `init_random` I added a check to ensure that `random_seed` is not still `""`. This was needed as, without a default value, it was not possible to pass `random_seed` with a parameter file or with `parameters` only;
- Fixed an additional test;

Time to pass to the examples in the tutorials and documentation:)

marcobortolami:
Benchmarks, tutorials and teaching material has been fixed

litebird_sim/simulations.py
mreineck: I'm very sure that this will give different RNGs on different tasks. The intent of `None` here seems to be to produce a truly random RNG, and the random devices on Linux are really good at this.
(Sorry for replying so late, I'm travelling.)

marcobortolami: ok nice:) thanks:)

marcobortolami:
I modified `test_simulation_random`: now it also checks that by passing an integer number as seed the results are reproducible, instead by passing `None` they are not.

marcobortolami:
For me, this is basically ready. I still have few questions / changes that I list here. Once we answer these, for me we can merge.

- [ ] **Change `litebird_sim` version.** This is a breaking change as all the scripts and notebooks instantiating a Simulation object (99% of them?) will fail. I'm not sure how to change the frameworks' version (like 0.10.0 --> 0.11.0). Is it something that is done at the PR level, for example with a commit?
- [ ] **Check figures that has changed.** Two pictures showing examples of reports in the documentation have been changed as now a new line writing the `random_seed` appears. How can I check the documentation as it appears from a browser before the PR is merged? Just to verify that the size of the pictures is ok.
- [ ] **Names.** When implementing this PR, I changed the name of the seed for the RNG from `seed` to `random_seed`. Although the usual name for a RNG seed is `seed`, I made this change as, in my opinion, its usage is more clear. To be more precise, when someone uses a seed is usually when creating a RNG, like `rng = np.random.random(seed)`. However, most of the times people will pass this parameter to the constructor of `Simulation`, like `sim = lbs.Simulation(random_seed=42)`. Here's my point: with `random_` it's more clear what that mandatory parameter is used for (random numbers). I'm aware that it's not customary, so if you don't like this I can change this name everywhere (easy task with a `grep`, just takes some time). I would also rename `sim.random` in `sim.rng` or `sim.random_number_generator`, but maybe it's too much;)
- [ ] **Update changelog.** Do this at the very last.

ziotom78:
Hi @marcobortolami , awesome work! I checked this PR and everything looks very nice.

Here are a few answers to your questions:

> Change litebird_sim version. This is a breaking change as all the scripts and notebooks instantiating a Simulation object (99% of them?) will fail. I'm not sure how to change the frameworks' version (like 0.10.0 --> 0.11.0). Is it something that is done at the PR level, for example with a commit?

No, this is usually tackled by a separate commit on the `master` branch. (It's better to keep version bumps separated from PR doing actual work.) We'll release a new version once this PR is merged.

We follow [semantic versioning](https://semver.org/), which means that until we hit version 1.0 we just bump the minor version every time we do a new release, without differentiating between breaking/non-breaking changes.

> Check figures that has changed. Two pictures showing examples of reports in the documentation have been changed as now a new line writing the random_seed appears. How can I check the documentation as it appears from a browser before the PR is merged? Just to verify that the size of the pictures is ok.

Just type

    make -C docs html

from the main directory. The `-C docs` flag tells Make to run within the `docs/` directory, and `html` is the name of the target. (Beware that you will likely need to install a few dependencies, most notably `sphinx`.) Once the command completes, just run

    firefox docs/build/html/index.html

(assuming you're using Firefox), and you'll be able to browse a local copy of the documentation. Every time you update the RST files, just re-run the `make` command and refresh the page.

> Names. When implementing this PR, I changed the name of the seed for the RNG from seed to random_seed. Although the usual name for a RNG seed is seed, I made this change as, in my opinion, its usage is more clear. To be more precise, when someone uses a seed is usually when creating a RNG, like rng = np.random.random(seed). However, most of the times people will pass this parameter to the constructor of Simulation, like sim = lbs.Simulation(random_seed=42). Here's my point: with random_ it's more clear what that mandatory parameter is used for (random numbers). I'm aware that it's not customary, so if you don't like this I can change this name everywhere (easy task with a grep, just takes some time). I would also rename sim.random in sim.rng or sim.random_number_generator, but maybe it's too much;)

I fully agree, `random_seed` is much clearer to read and clearly states the intent of the parameter. I like it!

Thank you for your hard work!

marcobortolami:
Thanks for the semantic versioning reference:) we'll release a new version after the PR is merged, understood.
I checked the documentation, thanks!
Ok for the names:)
I updated the changelog!

@ziotom78, thumb up here when I can merge!

ziotom78:
For me it's perfect! You can merge it whenever you want.

Thanks!

Issue #261, opened 2023-07-27
Error in reading observation when it does not have tod field


marcobortolami:
Hi:)
While working on the post-ptep e2e simulations, I noticed that if we try to read an observation saved without the `tod` field, I mean there is no `obs.tod`, then lines [543](https://github.com/litebird/litebird_sim/blob/eaef41c04a6725d207365a4b113ae24cac3ba49c/litebird_sim/io.py#L543) and [568-569](https://github.com/litebird/litebird_sim/blob/eaef41c04a6725d207365a4b113ae24cac3ba49c/litebird_sim/io.py#L568-L569) give an error as there is no `result.tod`.
I propose to change these lines as follows (this is a git diff output):
```
                 # All the fields must conform to the same shape as `Observation.tod`
-                assert result.tod.shape == hdf5_tod.shape
+                assert result.__getattribute__(tod_fields[0]).shape == hdf5_tod.shape
                 result.__setattr__(cur_field_name, hdf5_tod.astype(tod_dtype)[:])
 
         # If we arrive here, we must have read at least one TOD
@@ -565,8 +565,8 @@ def read_one_observation(
         if read_local_flags_if_present:
             flags = __find_flags(
                 inpf,
-                expected_num_of_dets=result.tod.shape[0],
-                expected_num_of_samples=result.tod.shape[1],
+                expected_num_of_dets=result.__getattribute__(tod_fields[0]).shape[0],
+                expected_num_of_samples=result.__getattribute__(tod_fields[0]).shape[1],
             )
```
I've tried with this modification and the code works. I can take care of the fix, I already created a local branch and I'm ready to commit in a PR:) What do you think?

ziotom78:
Hi Marco, thanks for having spotted this. No problems from my side, go on!

marcobortolami:
#262 solved this Issue.

PR #262, opened 2023-07-27
Fix error in reading observation when it does not have tod field

marcobortolami:
This PR solves #261.


marcobortolami:
Do you think we should also modify [test_io.py](https://github.com/litebird/litebird_sim/blob/master/test/test_io.py)? An easy peasy way would be to just change the names `tod` and `other_tod` to `tod1` and `tod2`, so the test is slightly more complex and tests also the case that was problematic, fixed with this PR.

marcobortolami:
I committed my idea of improvement for `test_io.py`:) I can revert the changes if you don't like them;)

ziotom78:
> I committed my idea of improvement for test_io.py:) I can revert the changes if you don't like them;)

Please keep it! I like it!

>     1. Why we define the [two tod field types](https://github.com/litebird/litebird_sim/blob/60dd81b08b5e43f9c13349f281d9e8972598a468/test/test_io.py#L94-L95) as float 64 and 32 but then we [assert](https://github.com/litebird/litebird_sim/blob/60dd81b08b5e43f9c13349f281d9e8972598a468/test/test_io.py#L232-L240) that they are both float 32? The test passes and I'm confused...

Because the function [`read_list_of_observations`](https://github.com/litebird/litebird_sim/blob/60dd81b08b5e43f9c13349f281d9e8972598a468/litebird_sim/io.py#L604) automatically converts every TOD it reads in the dtype specified by the parameter `tod_dtype`, and the default is `np.float32`.

>     2. [Here](https://github.com/litebird/litebird_sim/blob/60dd81b08b5e43f9c13349f281d9e8972598a468/test/test_io.py#L254-L283) we test the reading of an observation with and without mjd and gzip. However, not all the 4 combinations of the flags are present (`False`-`False`,`False`-`True`,`True`-`False`,`True`-`True`) because the `True`-`True` case is missing and the last test `test_quaternions_in_hdf5` is completely degenerate with `test_gzip_compression_in_obs`. Maybe one of the flags was mistakenly put to `False`? I propose to modify the last two test names as `test_gzip_compression_in_obs_mjd` and `test_gzip_compression_in_obs_no_mjd` with obvious setting of the flags.

No, the problem here is that `test_quaternions_in_hdf5` was a placeholder for a more complex test I have never had the chance to complete. (There are no quaternions in HDF5 so far…).

I wouldn't bother add the `True`-`True` case, as the code used to compress the file using GZip does not depend on the fact that MJD dates be used or not. The best thing is to just remove `test_quaternions_in_hdf5` altogether.

Thanks for having spotted these issues! From my side, this PR can be merged.

marcobortolami:
> Please keep it! I like it!

Ok:)

> Because the function [read_list_of_observations](https://github.com/litebird/litebird_sim/blob/60dd81b08b5e43f9c13349f281d9e8972598a468/litebird_sim/io.py#L604) automatically converts every TOD it reads in the dtype specified by the parameter tod_dtype, and the default is np.float32.

Oh, right, now I noticed the conversion!

> I wouldn't bother add the True-True case

Ok!

> The best thing is to just remove test_quaternions_in_hdf5 altogether.

Done!

Issue #263, opened 2023-08-01
Change python version in documentation + issue with python 3.9.17


marcobortolami:
Hi:) I create one Issue for, actually, two problems:

1. With #254 we dropped support for Python 3.7 and 3.8. However, in `installation.rst` (and in the slides that @nraffuzz and I prepared as tutorial) we say to the users to run `conda create -n lbs_env python=3.8`. These two files should be modified. I can take care of it. Which python version should we suggest?
2. I made a freshly new environment for `lbs` because I was using Python 3.8 before. I didn't know which Python 3.9.x to choose, so I chose the [latest one](https://www.python.org/doc/versions/), i.e. 3.9.17. However, the tests are not working (below a summary of what I did). Errors involve [`detectors.py`](https://github.com/litebird/litebird_sim/blob/master/litebird_sim/detectors.py). Should I use an older 3.9.x version? By solving point 1. above, also this could be solved.

```
conda create --name lbs_env python=3.9.17
[...]

conda activate lbs_env

git branch
* master
[...]

git pull
Already up-to-date.

pip install .
[...]

./bin/run_tests.sh
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running ``pip install "black[jupyter]"``
All done! ✨ 🍰 ✨
58 files would be left unchanged.
./litebird_sim/detectors.py:134:12: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
./litebird_sim/detectors.py:171:12: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
./litebird_sim/detectors.py:275:12: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
```
All the 3 errors refer to instructions where we do `type(whatever) != / == whatever_type`

ziotom78:
Regarding the first point, we can just ask for Python 3.9 and let Conda pick the latest bugfix release.

marcobortolami:
If I do `conda create --name lbs_env python=3.9`, then version 3.9.16 is used. The errors in point 2. above remain, as expected. Since it's just a minor change in the doc (8 --> 9), we could update `installation.rst` in #260 and solve all these problems with that PR:) I'll commit this now and review that PR tomorrow:)

ziotom78:
[...]

PR #265, opened 2023-08-05
Fix TOAST2 tests

ziotom78:
- [X] Make sure that TOAST2 is installed before testing the framework
- [x] Fix a few asserts in the TOAST2 tests


marcobortolami:
Is it trying to test toast with python 3.11? If it is the case, then I guess it is why it fails. All the tests are passing for me with python 3.9.16.

ziotom78:
This should fix #263 . It was due to the change from `assert np.allclose` to `np.testing.assert_allclose`, which are not 100% equivalent.

Issue #270, opened 2023-10-06
TypingError in destriper.py: cur_obs.destriper_weights is a float instead of array


nraffuzz:
While creating an Obsevation object (and filling it) I tried to apply the built-in destriper, this is the used code:

```
# fill the TOD
lbs.scan_map_in_observations(obs=obs,maps=input_map,pointings=pointings,input_map_in_galactic=True,component="tod")
# params for 1/f noise: fmin, net, fknee
obs.fmin_hz, obs.net_ukrts, obs.fknee_mhz = 1e-5, 2.2, 30
lbs.noise.add_noise_to_observations(obs=obs,noise_type='one_over_f',random=sim.random,scale=1,component="tod")                        

result = lbs.make_destriped_map(
    nside=nside, 
    obs=obs, 
    pointings=pointings, 
    params=lbs.DestriperParameters(), 
    components=['tod'],
    )
```

but I get the following error:
```
TypingError                               Traceback (most recent call last)
/home/nraffuzzi/mount_marconi/litebird/e2e-simulation/scripts/madam_vs_lbsdestriper.ipynb Cella 7 line 1
----> [1](vscode-notebook-cell://wsl%2Bubuntu/home/nraffuzzi/mount_marconi/litebird/e2e-simulation/scripts/madam_vs_lbsdestriper.ipynb#W6sdnNjb2RlLXJlbW90ZQ%3D%3D?line=0) result = lbs.make_destriped_map(
      [2](vscode-notebook-cell://wsl%2Bubuntu/home/nraffuzzi/mount_marconi/litebird/e2e-simulation/scripts/madam_vs_lbsdestriper.ipynb#W6sdnNjb2RlLXJlbW90ZQ%3D%3D?line=1)     nside=nside, 
      [3](vscode-notebook-cell://wsl%2Bubuntu/home/nraffuzzi/mount_marconi/litebird/e2e-simulation/scripts/madam_vs_lbsdestriper.ipynb#W6sdnNjb2RlLXJlbW90ZQ%3D%3D?line=2)     obs=obs, 
      [4](vscode-notebook-cell://wsl%2Bubuntu/home/nraffuzzi/mount_marconi/litebird/e2e-simulation/scripts/madam_vs_lbsdestriper.ipynb#W6sdnNjb2RlLXJlbW90ZQ%3D%3D?line=3)     pointings=pointings, 
[...]

PR #272, opened 2023-10-16
Solve typing error in destriper

nraffuzz:
This PR solves the issue #270.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # HEAD
 
+-   Solve typing error in destriper [#272](https://github.com/litebird/litebird_sim/pull/272)
+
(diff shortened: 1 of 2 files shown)


nraffuzz:
Since `weights` should be an array (field initialized by [get_map_making_weights](https://github.com/litebird/litebird_sim/blob/315a429ad020fab96776da35a43ca83147b61c1b/litebird_sim/mapmaking/common.py#L84-L106)), also `obs.net_ukrts` should be an array.
In order to handle this I added:
```
if isinstance(obs.net_ukrts, (float, int)):
    obs.net_ukrts = obs.net_ukrts * np.ones(obs.n_detectors)
```
inside `get_map_making_weights`.

I think this PR could be closed.

ziotom78:
Looks ok to me. Feel free to update the CHANGELOG and merge this PR!

Issue #285, opened 2024-01-14
numpy.float128 not supported on ARM 


paganol:
It seems that np.float128 is not available for ARM architectures. `io.py` fails when litebird_sim gets imported on macOS M3, I immagine is the same for M1 and M2

PR #291, opened 2024-02-24
Splits in mapmaker

paganol:
This PR contains a first attempt of introducing data and time splits in the mapmakers.

- [x] Implement the support for splits in `binner.py`
- [x] Fix tests for the binner 
- [x] Define splits 
- [x] Add high level functions for automatically loop over make_binned_map
- [x] Add documentation


litebird_sim/mapmaking/binner.py
```diff
@@ -95,38 +109,45 @@ def _accumulate_samples_and_build_nobs_matrix(
...
+                tod[idet, :], pix[idet, :], psi[idet, :], t_mask
+            ):
+                if cur_t_mask:
+                    cos_over_sigma = np.cos(2 * cur_psi) * inv_sigma
+                    sin_over_sigma = np.sin(2 * cur_psi) * inv_sigma
+                    info_pix = nobs_matrix[cur_pix_idx]
+
+                    info_pix[1, 0] += cur_sample * inv_sigma2
+                    info_pix[2, 0] += cur_sample * cos_over_sigma * inv_sigma
+                    info_pix[2, 1] += cur_sample * sin_over_sigma * inv_sigma
```
ziotom78, line 150: Small nitpick: with the new `if` the code gets one more nest level, but it is already quite nested with a sequence of `for`, `if`, `for`, `if`, and finally `if`. It's hard to see where the new `if d_mask[det]` ends…

A better solution would be to write the following statement at the top:

```python
if not d_mask[det]:
    continue
```

In this way it's crystal clear that if the `if` fails then the *entire* body of the `for` is skipped for this iteration.

litebird_sim/mapmaking/binner.py
```diff
@@ -259,7 +286,9 @@ def make_binned_map(
             to use for the output map
         components (list[str]): list of components to include in the map-making.
             The default is just to use the field ``tod`` of each
-            :class:`.Observation` object.
+            :class:`.Observation` object
+        detector_split (str): select the detector split to use in the map-making
+        time_split (str): select the time split to use in the map-making.
```
ziotom78, line 291: Once we have settled on a definitive list of splits we should list the accepted values in the docstrings, so that users do not need to resort to the manual to find out what to put in these arguments.

(Of course, this does not apply if we decide to use an enumeration here, as they are self-documenting.)

litebird_sim/mapmaking/common.py
```diff
@@ -418,3 +429,73 @@ def estimate_cond_number(
...
+            if isinstance(cur_obs.start_time_global, astropy.time.Time):
+                t_i = cur_obs.start_time_global.cxcsec
+            else:
+                t_i = cur_obs.start_time_global
+            time_mask.append(
+                ((cur_obs.get_times() - t_i) >= t_y2_sec)
+                * ((cur_obs.get_times() - t_i) < t_y3_sec)
+            )
+    else:
+        print("Time split not recognized!!!")
```
ziotom78, line 481: Better raise a `ValueError` exception:

```python
else:
    raise ValueError(f'Time split "{time_split}" not recognized')
```

test/test_simulations.py
```diff
@@ -238,7 +238,7 @@ def test_duration_units_in_parameter_file():
 
 
 def test_distribute_observation(tmp_path):
-    for dtype in (np.float16, np.float32, np.float64, np.float128):
+    for dtype in (np.float16, np.float32, np.float64, np.longdouble):
```
ziotom78, line 241: I wasn't aware of the existence of `np.longdouble`, which seems an alias for `np.float128`. Since we stopped supporting them because of issue #285 , perhaps it's better to remove them altogether…

litebird_sim/mapmaking/common.py
```diff
@@ -418,3 +429,73 @@ def estimate_cond_number(
...
+    obs_list: List[Observation],
+):
+    detector_mask = []
+
+    if detector_split == "full":
+        for cur_obs in obs_list:
+            detector_mask.append(np.ones(cur_obs.n_detectors, dtype=bool))
+    elif detector_split[0:6] == "wafer_":
+        for cur_obs in obs_list:
+            detector_mask.append(cur_obs.wafer == detector_split[6:9])
```
ziotom78, line 497: Small question: wouldn't be clearer to use just the name of the wafer without prepending `wafer_`? It would be probably easier for the user to remember too.

litebird_sim/mapmaking/common.py
```diff
@@ -418,3 +429,73 @@ def estimate_cond_number(
...
+    else:
+        print("Detector split not recognized!!!")
```
ziotom78, line 499: Better raise a `ValueError` here too.

ziotom78 (commented):
Thanks @paganol , it looks fine! I just commented on a few minor things.

litebird_sim/mapmaking/common.py
paganol: The tag `wafer_` is there for selecting the right attribute of `cur_obs` to check. In this example if `detector_split = "wafer_L01"` that piece of code compares `cur_obs.wafer` with `detector_split[6:9]`. If instead we want to select detectors per pixel type, we can use `detector_split = "pixtype_LP1"` and compare `cur_obs.pixtype` with `detector_split[8:11]` and so on... Does it seem reasonable?

litebird_sim/mapmaking/binner.py
paganol: Do you mean something like this?
```python
if not d_mask[det]:
    continue
else:
    inv_sigma = 1.0 / np.sqrt(weights[idet])
    ....
```

ziotom78: No, the `else` is redundant, the remaining part of the code can be un-indented.

litebird_sim/mapmaking/common.py
ziotom78: Ah, I see, ok.

litebird_sim/mapmaking/binner.py
paganol: Ah I see! right!

litebird_sim/mapmaking/binner.py
ggalloni: Hello @ziotom78, in this direction I implemented a function that checks the validity of the requested splits.
For example, it checks that a 1-year split fits in the requested observation duration, and so on.

I am still developing stuff of course, but the eventual documentation of this function should cover the valid splits etc, as you were suggesting.

ggalloni:
I enhanced the current loop over splits by moving it into the sims framework to access `write_healpix_map`. In such a way, we can save to disk each map to disk without having to allocate all of them. Still, there is a flag switching back to the dictionary if desired by the user. Also, I added some docs on the available splits within the function that check their validity.

Before committing, however, we may want to address an "issue" I am getting with ruff: running `ruff format .`, it does not complain of anything. Instead, when committing it shows extra requests to fix stuff related to formatting. I followed its suggestions for the previous commits, however, it started being quite annoying... Do you have any idea why it is doing this?

PR #308, opened 2024-03-19
Implement a time profiler

ziotom78:
This PR implements a simple time profiler and a high-level interface for the methods in the class `Simulation`.

Here is an example that shows how to use its low-level interface:

```python
import json
from litebird_sim import TimeProfiler, profile_list_to_speedscope
from time import sleep

perf_list = []  # type: list[TimeProfiler]

# First code block to profile
with TimeProfiler(name="function_A") as perf:
    sleep(1.0)
perf_list.append(perf)

# Second code block to profile
with TimeProfiler(name="function_B") as perf:
    sleep(2.0)
perf_list.append(perf)

# Save the profile measurements into "profile.json"
with open("profile.json", "wt") as out_f:
    json.dump(profile_list_to_speedscope(perf_list), out_f)

# Now go to https://www.speedscope.app/ and open "profile.json"
```

The high-level interface is automatically called by methods in the `Simulation` class. When a simulation ends and `Simulation.flush()` is called, a set of JSON files is created in the output directory. These files can be opened in the webapp <https://www.speedscope.app/>. This is how Speedscope looks if you open the file `profile.json` created by the script above:

speedscope-example

- [X] Implement the low-level interface
- [X] Implement a high-level interface for the `Simulation` class
- [X] Write tests
- [X] Write the documentation
- [x] Update the CHANGELOG


ziotom78:
I have implemented this because I am going to optimize the code used to compute the pointings, and I needed a simple way to monitor the time spent in different parts of the code.

Please test it and tell me if it looks ok: I would like to merge this in a few days, if no objections arise.

paganol:
Hi @ziotom78, I played at bit with the code. It looks fine to me. It will be quite useful in the future. Thanks!

PR #309, opened 2024-03-21
Data splits in destriper

ggalloni:
This PR is meant to implement the same data splits of the `binner` (see #291), but in the `destriper`.

- [x]  Implement same splits of #291
- [x]  Fix tests
- [x]  Add high-level function mimicking `make_binned_map_splits`
- [x] Add possibility to recycle baselines 
- [x]  Add documentation


ggalloni:
In this last commit, both binned and destriped maps out of the `destriper` result seem to be working.

Currently, only the single-split pipeline is in place, thus the `sim.make_destriped_map`.

ggalloni:
The loop over every split is implemented. 
However, in the current implementation, each split will search for its baselines.

It would be useful to give the possibility to "recycle" the baselines between splits.

ggalloni:
I added the recycling of baselines from the full splits. 

Apart from propagating the relevant attributes and changes, the main modification is that if baselines are passed to 'make_destriped_map' it does not iterate to find a new solution and instead uses the provided one. Also, I "stored" the info about the solution being recycled in the 'converged' attribute of the 'DestriperResult' object. Now, that can also be a string (not only a bool as before). This string says to the user that the baselines have been recycled from the full split and also reports the convergence of the original run. This might not be the best solution, let me know what you think eventually.

A "side-effect" of this modification is that now the user can pass a custom baseline solution obtained externally.

ggalloni:
I added a notebook doing some examples of data splits, for Binner and Destriper. 

If you can, have a look at its flow and let me know if it is OK.

After this, I think we can proceed with merging.

ggalloni:
Ciao @ziotom78 , questa PR penso sia mergiabile, se sei d'accordo.

Mi falliscono i test perché a quanto pare TOAST si sta rompendo su macos, ma non credo che la cosa dipenda dalla PR...

Con @paganol dicevamo che probabilmente la cosa migliore da fare è mergiare questa PR sul master e poi rimergiare il master dentro la #319 prima di toccare il destriper, che ne dici?

ziotom78:
> Hi @ziotom78, I think this PR is mergeable if you agree.
> 
> Some tests are failing because apparently TOAST is breaking on macos, but I doubt that depends on this PR...
> 
> I discussed the interplay of this PR with #319 with @paganol and the best thing to do seems to be: we merge this into the master and then we merge again the master in #319 before starting working on the destriper.
> 
> What do you think?

Great! I agree, let's do this.

PR #334, opened 2024-10-09
Distribution of detector blocks across MPI processes

anand-avinash:
This pull request is related to the discussion #325.

Here I have added a new method `_make_detector_blocks()` to the `Observation` class. It takes a list of the detectors and divides them into different groups as specified by `det_blocks_attributes` parameter passed to the `Observation` class. The group of detectors are saved in the variable `detector_blocks` of `Observation` class. For instance, with `det_blocks_attributes = ["wafer", "pixel"]`, the detector groups will be made in such a way that each group will have detectors with same wafer and pixel attributes.

Once the detectors are divided into the groups, it sets the number of detector blocks. The number of time blocks are computed accordingly (`n_blocks_time = comm.size // n_blocks_det`). Given the structure of the code base, this scheme was pretty easy to implement than I thought, and it doesn't breaks any other functionality.

On the other side, this implementation can lead to load imbalance in different MPI processes. Consider a simulation with 2 MPI processes and 4 detectors. If the detectors are ended up in two groups, with 3 and 1 elements respectively, one MPI process will have to work with 3 times more data than the other MPI process.

Another caveat is related to the `det_idx` member of `Observation` class. Previously, it corresponded to the global list of detectors passed to the `Observation` class. Now with the introduction of detector blocks, it won't be true anymore. For a simple solution, I have introduced a member variable `detectors_global`, that is the global list of detectors sorted according the detector groups. Now the `det_idx` will correspond to the indices of the detectors in `detectors_global`. But then we end up with three variables `detectors` (the list of detectors used to initialize the `Observation` class), `Observation.detector_blocks`, and `Observation.detectors_global` storing the same data in different manners.

After adding documentation, this pull request should become ready to be merged. But let me know if you have other solutions.

### DIFF
--- a/litebird_sim/distribute.py
+++ b/litebird_sim/distribute.py
@@ -84,6 +84,45 @@ def distribute_evenly(num_of_elements, num_of_groups):
     return result
 
 
+def distribute_detector_blocks(detector_blocks):
+    """Similar to the function :func:`distribute_evenly()`, this function returns the named-tuples of the starting index of the detectors in a group
+    with respect to the global list of detectors and the number of detectors
+    in the group. Unlike the :func:`distribute_evenly()`, this function simply
+    uses the detector groups given in `detector_blocks` attribute.
+
+    Example:
+        Following the example given in
+        :meth:`litebird_sim.Observation._make_detector_blocks()`,
+        `distribute_detector_blocks()` will return
+
+        ```
+        [
+            Span(start_idx=0, num_of_elements=2),
+            Span(start_idx=2, num_of_elements=2),
+            Span(start_idx=4, num_of_elements=1),
+        ]
+        ```
+
+    Args:
+        detector_blocks (dict): The detector block object. See :meth:`litebird_sim.Observation._make_detector_blocks()`.
+
+    Returns:
+        A list of 2-elements named-tuples containing (1) the starting index of
+        the detectors of the block with respect to the flatten list of entire
+        detector blocks and (2) the number of elements in the detector block.
+    """
+    cur_position = 0
+    prev_length = 0
+    result = []
+    for key in detector_blocks:
+        cur_length = len(detector_blocks[key])
+        cur_position += prev_length
+        prev_length = cur_length
+        result.append(Span(start_idx=cur_position, num_of_elements=cur_length))
+
+    return result
+
+
 # The following implementation of the painter's partition problem is
 # heavily inspired by the code at
 # https://www.geeksforgeeks.org/painters-partition-problem-set-2/?ref=rp
--- a/litebird_sim/observations.py
+++ b/litebird_sim/observations.py
@@ -7,8 +7,10 @@ import astropy.time
 import numpy as np
 import numpy.typing as npt
 
+from collections import defaultdict
+
 from .coordinates import DEFAULT_TIME_SCALE
-from .distribute import distribute_evenly
+from .distribute import distribute_evenly, distribute_detector_blocks
 
 
 @dataclass
@@ -80,11 +82,19 @@ class Observation:
 
         sampling_rate_hz (float): The sampling frequency, in Hertz.
 
+        det_blocks_attributes (list of strings): The list of detector attributes that
+            will be used to divide detector axis of the tod (and all their attributes).
+            For example, with ``det_blocks_attributes = ["wafer", "pixel"]``, the
+            detectors will be divided into the blocks such that all detectors in a
+            block will have same ``wafer`` and ``pixel`` attribute.
+
         n_blocks_det (int): divide the detector axis of the tod (and all the
-            arrays of detector attributes) in `n_blocks_det` blocks
+            arrays of detector attributes) in `n_blocks_det` blocks. Will be ignored
+            if ``det_blocks_attributes`` is not `None`.
 
         n_blocks_time (int): divide the time axis of the tod in
-            `n_blocks_time` blocks
+            `n_blocks_time` blocks. Will be ignored if ``det_blocks_attributes``
+            is not `None`.
 
         comm: either `None` (do not use MPI) or a MPI communicator
             object, like `mpi4py.MPI.COMM_WORLD`. Its size is required to be at
@@ -103,6 +113,7 @@ class Observation:
         sampling_rate_hz: float,
         allocate_tod=True,
         tods=None,
+        det_blocks_attributes: Union[List[str], None] = None,
         n_blocks_det=1,
         n_blocks_time=1,
         comm=None,
@@ -123,6 +134,10 @@ class Observation:
             delta = 1.0 / sampling_rate_hz
         self.end_time_global = start_time_global + n_samples_global * delta
 
+        self._sampling_rate_hz = sampling_rate_hz
+        self._det_blocks_attributes = det_blocks_attributes
+        self.detector_blocks = None
+
         if isinstance(detectors, int):
             self._n_detectors_global = detectors
         else:
@@ -131,9 +146,12 @@ class Observation:
             else:
                 self._n_detectors_global = len(detectors)
 
-        self._sampling_rate_hz = sampling_rate_hz
+            if self._det_blocks_attributes is not None and comm.size > 1:
+                n_blocks_det, n_blocks_time = self._make_detector_blocks(
+                    detectors, comm
+                )
 
-        # Neme of the attributes that store an array with the value of a
+        # Name of the attributes that store an array with the value of a
         # property for each of the (local) detectors
         self._attr_det_names = []
         self._check_blocks(n_blocks_det, n_blocks_time)
@@ -159,8 +177,17 @@ class Observation:
                 setattr(self, cur_tod.name, None)
 
         self.setattr_det_global("det_idx", np.arange(self._n_detectors_global), root)
+
+        self.detectors_global = []
+
+        if self.detector_blocks is not None:
+            for key in self.detector_blocks:
+                self.detectors_global += self.detector_blocks[key]
+        else:
+            self.detectors_global = detectors
+
         if not isinstance(detectors, int):
-            self._set_attributes_from_list_of_dict(detectors, root)
+            self._set_attributes_from_list_of_dict(self.detectors_global, root)
 
         (
             self.start_time,
@@ -203,7 +230,7 @@ class Observation:
         return self.start_time_global + start * delta, start, num
 
     def _set_attributes_from_list_of_dict(self, list_of_dict, root):
-        assert len(list_of_dict) == self.n_detectors_global
+        np.testing.assert_equal(len(list_of_dict), self.n_detectors_global)
 
         # Turn list of dict into dict of arrays
         if not self.comm or self.comm.rank == root:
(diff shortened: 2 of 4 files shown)


litebird_sim/distribute.py
```diff
@@ -84,6 +84,45 @@ def distribute_evenly(num_of_elements, num_of_groups):
     return result
 
 
+def distribute_detector_blocks(detector_blocks):
+    """Similar to the function :func:`distribute_evenly()`, this function returns the named-tuples of the starting index of the detectors in a group
+    with respect to the global list of detectors and the number of detectors
+    in the group. Unlike the :func:`distribute_evenly()`, this function simply
+    uses the detector groups given in `detector_blocks` attribute.
```
ziotom78, line 91: Clearer:

> Similar to the distribute_evenly() function, this function returns a list of named tuples with fields `start_idx` (the starting index of the detector in a group within the global list of detectors) and `num_of_elements` (the number of detectors in the group). Unlike distribute_evenly(), this function uses the detector groups given in the `detector_blocks` attribute.

litebird_sim/observations.py
```diff
@@ -80,11 +82,19 @@ class Observation:
 
         sampling_rate_hz (float): The sampling frequency, in Hertz.
 
+        det_blocks_attributes (list of strings): The list of detector attributes that
+            will be used to divide detector axis of the tod (and all their attributes).
+            For example, with ``det_blocks_attributes = ["wafer", "pixel"]``, the
+            detectors will be divided into the blocks such that all detectors in a
+            block will have same ``wafer`` and ``pixel`` attribute.
```
ziotom78, line 89: Fix the grammar:

> The list of detector attributes that will be used to divide the detector axis of the tod matrix and all its attributes. For example, with `det_blocks_attributes = ["wafer", "pixel"]`, the detectors will be divided into blocks such that all detectors in a block will have the same ``wafer`` and ``pixel`` attributes.

litebird_sim/observations.py
```diff
@@ -80,11 +82,19 @@ class Observation:
...
+
         n_blocks_det (int): divide the detector axis of the tod (and all the
-            arrays of detector attributes) in `n_blocks_det` blocks
+            arrays of detector attributes) in `n_blocks_det` blocks. Will be ignored
+            if ``det_blocks_attributes`` is not `None`.
```
ziotom78, line 93: “It will be ignored…”

litebird_sim/observations.py
```diff
@@ -80,11 +82,19 @@ class Observation:
...
 
         n_blocks_time (int): divide the time axis of the tod in
-            `n_blocks_time` blocks
+            `n_blocks_time` blocks. Will be ignored if ``det_blocks_attributes``
+            is not `None`.
```
ziotom78, line 97: “It will be ignored…”

litebird_sim/observations.py
```diff
@@ -273,10 +300,86 @@ def n_blocks_time(self):
...
+        detector blocks is set to be the total number of detector groups
+        whereas the number of time blocks is computed using the number of
+        detector blocks and the size of `comm` communicator.
+
+        The blocks of detectors are stored in `self.detector_blocks`. It is a
+        dictionary object with the tuple of `self._det_blocks_attributes` values
+        as dictionary keys and the list of detectors corresponding to the key
+        as the dictionary value. This dictionary is sorted so that that the
+        group with largest number of detectors comes first and the one with
+        the least number of detectors, comes last.
```
ziotom78, line 316: Better:

> This function distributes the detectors in groups such that each group has the same set of attributes specified by the strings in `self._det_block_attributes`. Once the groups are made, the number of detector blocks is set to be the total number of detector groups, whereas the number of time blocks is computed using the number of detector blocks and the size of the `comm` communicator.
The detector blocks are stored in `self.detector_blocks`. This dictionary object has the tuple of `self._det_blocks_attributes` values as dictionary keys and the list of detectors corresponding to the key as dictionary values. This dictionary is sorted so that the group with the largest number of detectors comes first and the one with the fewest detectors comes last.

litebird_sim/simulations.py
```diff
@@ -922,7 +923,12 @@ def create_observations(
         simulating 10 detectors and you specify ``n_blocks_det=5``,
         this means that each observation will handle ``10 / 5 = 2``
         detectors. The default is that *all* the detectors be kept
-        together (``n_blocks_det=1``).
+        together (``n_blocks_det=1``). On the other hand, the parameter
+        `det_blocks_attributes` specifies the list of detector attributes
+        to be used to create the groups of detectors. For example, with
+        ``det_blocks_attributes = ["wafer", "pixel"]``, the detectors will
+        be divided into the groups such that all detectors in a group will
+        have same ``wafer`` and ``pixel`` attribute.
```
ziotom78, line 931: Better:

> On the other hand, the parameter `det_blocks_attributes` specifies the list of detector attributes to create groups of detectors. For example, with ``det_blocks_attributes = ["wafer", "pixel"]``, the detectors will be divided into groups such that all detectors in a group will have the same ``wafer`` and ``pixel``attributes.

ziotom78 (commented):
Thanks for having fixed the comments!

anand-avinash:
I believe this PR is ready to be merged, but there are a few minor issues that need discussion. The implementation of attribute-based detector blocks works well in the full simulation pipeline when `comm.size` equals `n_blocks_det * n_blocks_time`. However, we encounter errors when `comm.size` is larger, as this leaves some MPI processes unused. The `Observation` classes on the unused processes are still being queried at several places, leading to the errors. When we provide `det_blocks_attributes`, the `n_blocks_time` is computed during runtime using the value of `comm.size` (`n_blocks_time = comm.size // n_blocks_det`). So we cannot avoid this situation unless we do the simulations with a well-balanced set of detectors.

Nonetheless, I tried to resolve some of the error which I am documenting here.

The first error I encountered was:
```
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/BrahMap/test_examples/cross_talk/detector_blocks_full_example.py", line 127, in <module>
    binned_maps = lbs.make_binned_map(nside, sim.observations)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/binner.py", line 325, in make_binned_map
    detector_mask_list = _build_mask_detector_split(detector_split, obs_list)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/common.py", line 515, in _build_mask_detector_split
    detector_mask.append(np.ones(cur_obs.n_detectors, dtype=bool))
                                 ^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/observations.py", line 212, in n_detectors
    return len(self.det_idx)
           ^^^^^^^^^^^^^^^^^
TypeError: object of type 'NoneType' has no len()
```
which I solved simply by adding the following:
https://github.com/litebird/litebird_sim/blob/a941fd0f7fbf72f441c17dd4e2edfe732fb9909b/litebird_sim/observations.py#L209-L212

---

The next error I encountered was
```
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/BrahMap/test_examples/cross_talk/detector_blocks_full_example.py", line 127, in <module>
    binned_maps = lbs.make_binned_map(nside, sim.observations)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/binner.py", line 329, in make_binned_map
    nobs_matrix = _build_nobs_matrix(
                  ^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/binner.py", line 209, in _build_nobs_matrix
    cur_weights = get_map_making_weights(cur_obs, check=True)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/common.py", line 107, in get_map_making_weights
    weights = observations.sampling_rate_hz * observations.net_ukrts**2
                                              ~~~~~~~~~~~~~~~~~~~~~~^^~
TypeError: unsupported operand type(s) for ** or pow(): 'NoneType' and 'int'
```
It appears because `setattr_det_global()` sets the detector attributes to `None` on the processes that are unused. I was able to fix it by setting the numerical attributes to `0` instead of `None`:
https://github.com/litebird/litebird_sim/blob/a941fd0f7fbf72f441c17dd4e2edfe732fb9909b/litebird_sim/observations.py#L677-L682

---
The next error I encountered is
```
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/BrahMap/test_examples/cross_talk/detector_blocks_full_example.py", line 127, in <module>
    binned_maps = lbs.make_binned_map(nside, sim.observations)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/binner.py", line 329, in make_binned_map
    nobs_matrix = _build_nobs_matrix(
                  ^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/binner.py", line 224, in _build_nobs_matrix
    pixidx_all, polang_all = _compute_pixel_indices(
                             ^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/common.py", line 222, in _compute_pixel_indices
    del curr_pointings_det
        ^^^^^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'curr_pointings_det' where it is not associated with a value
```
It can be solved by replacing line 222 of `litebird_sim/common.py` by following:
```
        try:
            del curr_pointings_det
        except UnboundLocalError:
            pass
```

---

Next one I encountered while calling the destriper:

```
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/destriper.py", line 1619, in make_destriped_map
    len(params.samples_per_baseline)
TypeError: object of type 'int' has no len()

During handling of the above exception, another exception occurred:

  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/BrahMap/test_examples/cross_talk/detector_blocks_full_example.py", line 131, in <module>
    destriped_maps = lbs.make_destriped_map(nside, sim.observations)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/destriper.py", line 1630, in make_destriped_map
    baseline_lengths_list = [
                            ^
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/destriper.py", line 1631, in <listcomp>
    split_items_evenly(
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/lbsim_v0.13.0/litebird_sim/mapmaking/destriper.py", line 101, in split_items_evenly
    assert sub_n < n, "sub_n={0} is not smaller than n={1}".format(sub_n, n)
           ^^^^^^^^^
AssertionError: sub_n=100 is not smaller than n=0
```
This one seems little non-trivial to solve.

At the end we may face errors elsewhere as well that I have not tested. And it brings me to the question: should we worry about fixing this issue or should we leave it to the user to supply a well-balanced set of detectors?

ziotom78:
> At the end we may face errors elsewhere as well that I have not tested. And it brings me to the question: should we worry about fixing this issue or should we leave it to the user to supply a well-balanced set of detectors?

The simplest solution would probably be to check in `Simulation.__init__` and crash if the number of MPI processes is too large. I’m not sure that this is the best option, we should ask the opinion of people of the Simulation Production Team.

anand-avinash:
As we discussed in the telecon, I have now added a new subcommunicator that excludes the processes from `MPI.COMM_WORLD` where there are no detectors. It includes all the processes for which `rank < n_blocks_det*n_blocks_time`. The new subcommunicator is encapsulated in a class instance (`litebird_sim.mpi.comm_grid`) defined in `mpi.py`, with explicitly defined setter function so that it cannot be overwritten accidently. The new subcommunicator can be accessed with `litebird_sim.mpi.comm_grid.COMM_OBS_GRID`. By default, this subcommnicator is equal to `MPI.COMM_WORLD` unless the observations are defined. Depending on the number of detector blocks and time blocks, the value of the subcommunicator is updated.

For the processes that do not belong to this new sub-communicator, the sub-communicator object points to a NULL communicator (`MPI.COMM_NULL`). When using `litebird_sim.mpi.COMM_OBS_GRID`, one must check if it is not same as the NULL communicator. The general NULL communicator can be accessed with `litebird_sim.mpi.comm_grid.COMM_NULL`.

To resolve the issues I reported previously, I have replaced the global communicator with the subcommunicator at various places. However, there might be some other places where this replacement is necessary, for example `io.py`, `hwp_sys.py`, etc. We can update them if someone encounters an error in future.

@ziotom78 @paganol please let me know if we can proceed with this solution.

litebird_sim/mpi.py
```diff
@@ -22,10 +22,47 @@ class _SerialMpiCommunicator:
...
+    def _set_null_comm(self, comm_null):
+        self._MPI_COMM_NULL = comm_null
+
+
 #: Global variable equal either to `mpi4py.MPI.COMM_WORLD` or a object
 #: that defines the member variables `rank = 0` and `size = 1`.
 MPI_COMM_WORLD = _SerialMpiCommunicator()
 
+comm_grid = _GridCommClass()
+
```
ziotom78, line 65: Being a global variable that is not expected to be reassigned, I would name it `MPI_COMM_GRID`, in analogy with `MPI_COMM_WORLD`, `MPI_ENABLED`, etc.

ziotom78 (changes requested):
I like it! Nice work!

litebird_sim/mpi.py
ziotom78: Refactoring a variable name is trivial if you use an advanced IDE like PyCharm. If you don't, let me know, and I'll rename it. (It’s just a matter of one mouse click.)

anand-avinash: > Being a global variable that is not expected to be reassigned, I would name it `MPI_COMM_GRID`, in analogy with `MPI_COMM_WORLD`, `MPI_ENABLED`, etc.

Do you mean removing the `_GridCommClass` altogether and instead simply use `MPI_COMM_GRID`?

ziotom78: > > Being a global variable that is not expected to be reassigned, I would name it `MPI_COMM_GRID`, in analogy with `MPI_COMM_WORLD`, `MPI_ENABLED`, etc.
> 
> Do you mean removing the `_GridCommClass` altogether and instead simply use `MPI_COMM_GRID`?

No, it's just a matter of renaming `comm_grid`, as in this way it will match the naming convention of the other global variables in the `mpi` module.

ziotom78:
My gosh, thank you so much for having spotted this, @anand-avinash ! It was added four years ago but never documented.

> It might be useful to make the sub-communicators for detector and time blocks the members of Observation class. The detector block sub-communicator can be used to pick up all the ranks where the TOD chunks of a given detector reside. Similarly the time block sub-communicator can be used to pick all the ranks where TOD chunks of a given time interval is located.

Absolutely, I agree!

PR #337, opened 2024-10-30
Add Mueller convolver, take 2

mreineck:
This replaces #170, which seems to have gone too stale.


mreineck:
This is now integrated into #338, so we can close it.

Issue #183, opened 2022-07-21
Beam convolution


giuspugl:
Hello, 
as discussed am posting the links related to the current implementation of the Convolution Operators in toast3, 

https://github.com/hpc4cmb/toast/blob/c964a2586575e88144eb80206ca869d2ba012ab6/src/toast/ops/conviqt.py#L696

and the test routines used to validate the implementation  
https://github.com/hpc4cmb/toast/blob/c964a2586575e88144eb80206ca869d2ba012ab6/src/toast/tests/ops_sim_tod_conviqt.py#L460

mreineck:
After way too much procrastinating (sorry!) I'll try to pick this up.

To make sure I understand things right: the task is to get `mueller_convolver` integrated into the `litebird_sim` API, correct?

If that is the case, our best starting point is probably @okayamayu8's work, who has been running HWP simulations with a modified version of `litebird_sim` recently.

@okayamayu8, would it be OK for you to share your code for future integration, perhaps as part of #170?

okayamayu8:
Sure.
As for the code I shared, you can use it as you like.

mreineck:
@okayamayu8 , thank you very much!

I'm sorry that I'm probably missing something, but in the Julia notebook you shared I don't actually see any direct usage of litebird_sim or mueller_convolver. My impression was that you were probably working on a Python code that integrated mueller_convolver with litebird_sim and ran on supercomputer nodes. If that is indeed the case, it would be great to use this for this issue. If not, we probably have to implement this from scratch ... unfortunately I don't really feel qualified for this :-/

okayamayu8:
Sorry, I didn't understand exactly what you wanted to do.
That note uses random pointing as a validation of the mueller convolver.
So it is not working with litebird_sim.
And for mapmake, which we have been discussing recently, we were using a combination of litebird_sim and other external functions (written in julia).
So I had escaped from Python code.
So I am aware that we need to create from scratch a script that is complete with only litebird_sim and mueller convolver.

ziotom78:
Hi @mreineck , looking at the TOAST code, it seems that you need pointings, TODs, and synthetic maps; they are described at the links <https://litebird-sim.readthedocs.io/en/latest/observations.html> and <https://litebird-sim.readthedocs.io/en/latest/sky_maps.html>. Here are a few notes:

[...]

PR #338, opened 2024-10-30
Add 4pi beam convolution

mreineck:
This aims to add 4pi convolution with non-axisymmetric beams (very similar to the "conviqt" functionality in TOAST, but more accurate and hopefully also quicker).
Currently this cannot be fully implemented because the code requires
- a_lm of the sky model maps (probably solved by #306, I need to check how I can access them).
- a_lm of the beams. Those will be in the IMO at some point, but have not yet been integrated.

It should be possible to load the beams from files for testing purposes, using `healpy.read_alm`, but I'm not sure how to provide the corresponding file names in the input parameters and how they would be passed to the beam convolution module.
Unfortunately I'm not very familiar with the simulation framework and the really large set of tools it uses. I'm happy to help tweaking and debugging the underlying algorithms, but I need help from experts with the high-level integration.


mreineck:
I must say that some of the pre-commit hooks are driving me quite crazy ...
I attempt to add the following lines to `litebid_sim/__init__.py`:

```Python
from .beam_convolution import (
    add_convolved_sky_to_observations,
)
```
This is what I'm trying to do:
```
(lbsim) martin@marvin:~/codes/litebird_sim$ git status
On branch beam_comvolution_2
Your branch is up to date with 'origin/beam_comvolution_2'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   litebird_sim/__init__.py

no changes added to commit (use "git add" and/or "git commit -a")
(lbsim) martin@marvin:~/codes/litebird_sim$ git diff
diff --git a/litebird_sim/__init__.py b/litebird_sim/__init__.py
index b4a8aa6..1985a0c 100644
--- a/litebird_sim/__init__.py
+++ b/litebird_sim/__init__.py
@@ -150,6 +150,9 @@ from .spacecraft import (
     SpacecraftOrbit,
     SpacecraftPositionAndVelocity,
 )
+from .beam_convolution import (
+    add_convolved_sky_to_observations,
+)
 from .version import __author__, __version__
 
 # Check if the TOAST2 mapmaker is available
```

So I have the mentioned changes in my tree.
Now I run
```
(lbsim) martin@marvin:~/codes/litebird_sim$ git commit -a -m "add method to __init__.py"
ruff.....................................................................Failed
- hook id: ruff
- files were modified by this hook

Found 1 error (1 fixed, 0 remaining).

ruff-format..............................................................Passed
```

If I now look at the status, all changes are gone!
```
(lbsim) martin@marvin:~/codes/litebird_sim$ git status
On branch beam_comvolution_2
Your branch is up to date with 'origin/beam_comvolution_2'.

nothing to commit, working tree clean
```

As you can see, the pre-commit hooks somehow remove my changes from `__init__.py`! What is going wrong here?

paganol:
Hi @mreineck, I can help in the high level integration. I can work on this branch and prepare a coherent `lbs` interface, what do you think? 
Regarding the pre-commit, I cannot help, I never had this kind of problem

mreineck:
Thank you so much, @paganol, this is a great relief for me!

Concerning the precommit hook, I managed to get the crucial information of of `ruff` by ading `-v`:
```
(lbsim) martin@marvin:~/codes/litebird_sim$ ruff litebird_sim/__init__.py 
litebird_sim/__init__.py:22:5: F401 [*] `.beam_convolution.add_convolved_sky_to_observations` imported but unused
Found 1 error.
[*] 1 fixable with the `--fix` option.
```

So I need to add the new function to `__all__` further down in the file. Makes a lot of sense, but I really don't feel that silently removing the `import` was helpful to understand the problem. Maybe we shouldn't use `--fix` and just let `ruff` complain about things (and explain them), until everything has been fixed manually.

mreineck:
I wonder whether we should merge #337 wit this PR, then we can add the signal modulation by the HWP as well (optionally, of course!). The only additional required quantity for this is the time-dependent HWP angle.

mreineck:
I know that the PR doesn't look like much, but apart from obtaining the missing input data (which, I admit, may be complicated to achieve!), it is pretty much complete. All the numerical details are handled by `ducc`.

mreineck:
Thank you, @paganol, your interfacing work looks great to me!
I see that some error messages still refer to `scan_map_in_observations`; I'm happy to update them, unless there is some deeper connection to `scan_map_in_observations` that I don't see at the moment.
I'll also copy over `mueller_convolver.py` from #337 and close #337 afterwards; having these two changes in a single PR makes much more sense!

paganol:
Hi @mreineck, sorry I forgot to update the error messages! No deeper connection 😅... Of course if we find better names for the lower level functions we can change them. About merging the PRs, I totally agree with you.

mreineck:
Stupid question: is there a place in `litebird_sim` where I can put `mueller_convolver.py`, so that it is not reformatted by `ruff`? Strictly speaking this piece of code is maintained somewhere else (although it doesn't have an "official" place yet), and reformatting (and especially the re-naming of variables that `ruff` requests) will be quite a maintenance headache.
Should we have an `external/` subdirectory?

ziotom78:
> Stupid question: is there a place in `litebird_sim` where I can put `mueller_convolver.py`, so that it is not reformatted by `ruff`? Strictly speaking this piece of code is maintained somewhere else (although it doesn't have an "official" place yet), and reformatting (and especially the re-naming of variables that `ruff` requests) will be quite a maintenance headache. Should we have an `external/` subdirectory?

Would [`#fmt: off`](https://docs.astral.sh/ruff/formatter/#format-suppression) at the beginning of the file be ok, or do you want the file to be left completely untouched?

mreineck:
No, that's perfectly fine, thanks for pointing this out!

paganol:
Hi @mreineck, for matching the latest changes in the pointing API (PR #340) this line 
https://github.com/litebird/litebird_sim/blob/b671b68fa014d0373abae410ed0d4ac0f2e1dede/litebird_sim/beam_convolution.py#L65
should be removed.

mreineck:
Thanks for te tip, I'll fix this!
I'm travelling right now, but I hope to get back to this PR next week.

paganol:
Hi @ziotom78 @mreineck, I managed to get the convolver kind of working. For now only the case including the hwp and in double precision works, and in the other cases the interface with ducc0 returns errors. I'm trying to fix these remaining issues.

Anyway, this interface now works:
```
Convparams = lbs.BeamConvolutionParameters(
    lmax = lmax,
    kmax = lmax-4,
    single_precision = False,
    epsilon = 1e-7,    
)
sim.convolve_sky(sky_alms=alms,
                 beam_alms=blms,
                 convolution_params=Convparams,
                 input_sky_alms_in_galactic=True,
                )
```
And for Gaussian beams returns a reasonable convolved map, at least for a first attempt.. 
Here, it is compared with an already convolved scanned map.
Convolution

Here, for me what we should do as as next steps:  
- [ ] Fix the remaining interfaces
- [ ] Validate the orientation
- [ ] Add unit tests
- [ ] Adjust docstrings
- [ ] Add documentation
- [ ] Add example notebooks

paganol:
Now the interfaces work for both single and double precision. Still we need to agree on how to handle `lmax` and `kmax`, the convolver rises errors if the size of alms and blms does not match these parameters.

mreineck:
Thanks a lot for all the work you are doing, this look great!

Concerning `lmax` and `kmax`: they have to match what is in the alms and blms, because these are simple 1D structures (per a_lm component) from which their own `lmax` and `mmax` cannot be deduced unambiguously, at least not both at the same time.

`healpy` has a `resize_alm` function which could be used here.

paganol:
Hi @mreineck, exactly, if mmax $\neq$ lmax, it's impossible to get both at the same time. I was wondering what is the best strategy for the interface. Should we pass the lmax and mmax of the blms together with blms, wrapping everything in a dictionary, and then let the interface do the proper resize?

mreineck:
I don't have any preference here, to be honest. At the lowest level, where I usually work, I would expect both `slm` and `blm` to be appropriately resized to match the specified `lmax` and `kmax`; but at this intermediate level, your suggestion is perfectly fine and should not add a lot of overhead in terms of CPU time and memory.

(Perhaps the cleanest thing would have been to add mandatory `lmax` and `mmax` parameters to the `healpy` alm type ... but hindsight is always 20/20...)

mreineck:
The required `ducc` change is now on branch `flexible_sht_types` and ready to merge.
However I fear that it will only be "officially" be usable by LBS if I make a new `ducc` release, correct? That would be a bit awkward, since I made the last one only ten days ago ...

ziotom78:
> The required `ducc` change is now on branch `flexible_sht_types` and ready to merge. However I fear that it will only be "officially" be usable by LBS if I make a new `ducc` release, correct? That would be a bit awkward, since I made the last one only ten days ago ...

Unfortunately, that’s right 🙁. We install `ducc` (and any other dependency) through PyPI… Perhaps we can wait a bit until we implement a few tests? In this case, if we find a few other things that might be improved in `ducc`, you can accumulate them in just one new release.

mreineck:
Sounds like a perfect solution to me!

litebird_sim/beam_convolution.py
```diff
@@ -0,0 +1,313 @@
...
+        if hwp is None:
+            if hasattr(cur_obs, "hwp_angle"):
+                hwp_angle = cur_obs.hwp_angle
+            else:
+                hwp_angle = None
+        else:
+            if type(cur_ptg) is np.ndarray:
+                hwp_angle = get_hwp_angle(cur_obs, hwp)
+            else:
+                logging.warning(
```
ziotom78, line 298: In this `else` block, `hwp_angle` is not initialized and will cause an error in the call to `add_convolved_sky` later

ziotom78:
Hi @paganol , I am reasonably happy with the `SphericalHarmonics` class. I started updating `beam_convolution.py` to support this new class, but I stopped because it’s not 100% clear how to fix two places in the code (marked with a `TODO`) comment.

I have already started working on a test for the beam orientation, where I simulate a simple scanning strategy that orbits around a point source (with no drift nor precession) and checks for the correctness of the convolution in a semi-analytical way. I will complete it once `beam_convolution.py` is fixed.

paganol:
Hi @ziotom78 and @mreineck, I fixed the interface. I made the following changes:
- In `mbs.py`, I corrected the interface with `map2alm`. It was assuming that `map2alm` was returning a tuple, but it returns always a numpy array. Now this is fixed. 
- I added a method to `SphericalHarmonics`, called `resize_alm`, for resizing the alm object.
- I used this resize function in `add_convolved_sky_to_one_detector`, now the size of the input alms is adjusted and it matches automatically the needs of `MuellerConvolver`

There are few things that need to be fixed/checked:
- The output of `gauss_beam_to_alm` does not match the output of `grasp2alm`, there is for sure a factor 2 floating, and probably also some ell dependent factor, in case of symmetric Gaussian beam. Also we might define some defaults for it (like symmetric gaussian). I can work on it during PO/F2F with the help of @yusuke-takase.
- I interfaced the nthreads variable with `OMP_NUM_THREADS` and with `Simulation.numba_threads`. This needs to be checked and, if everything is fine, we might change the name of `numba_threads` to `num_threads` or similar.. 
- The convolution seems to work, at least for temperature and polarization in absence of HWP. With the HWP polarization maps make no sense.. There is something not working in `MuellerConvolver`.. I could not find it yet.. 

After these points I think we can procede with the validation using an elliptical beam.

paganol:
Sorry, I forgot one thing.. probably whole interface in `beam_convolution.py` could be optimized a bit.. but for now it does what it's supposed to do..

ziotom78:
>     * The output of `gauss_beam_to_alm` does not match the output of `grasp2alm`, there is for sure a factor 2 floating, and probably also some ell dependent factor, in case of symmetric Gaussian beam. Also we might define some defaults for it (like symmetric gaussian). I can work on it during PO/F2F with the help of @yusuke-takase.

Thanks for looking at this! The numbers I used in the test `test_gaussbeam_elliptical()` (file `test/test_beam_synthesis.py`) were produced using the Fortran script available at this link: <https://gist.github.com/ziotom78/550fcfe3d488e5d1fcb5829a4db45ba8>. Perhaps it could be useful for you and Yusuke to look at it and see what’s wrong.

mreineck:
Thank you so much for all the work, @paganol! Everything I've seen so far looks great to me; unfortunately I don't have time for digging very deep at the moment, since I have a visitor this week and will be travelling next week :-/

yusuke-takase:
Hi @paganol, @ziotom78 

I checked grasp2alm and litebird_sim their beam functions. 
The mismatch that Luca observed is caused by a convention of multipole and normalization factor of the gaussian beam's transfer function. 

lbs and levelS are using ([lbs line](https://github.com/litebird/litebird_sim/blob/beam_comvolution_2/litebird_sim/beam_synthesis.py#L91), [levelS line](https://github.com/zonca/planck-levelS/blob/master/Beam/beam_alm.f90#150)):
$b_\ell = \sqrt{\frac{2 \ell + 1}{4 \pi}}\exp{(-\frac{1}2 \sigma^2 \ell^2)}$
and grasp2alm is using ([line](https://github.com/yusuke-takase/grasp2alm/blob/master/grasp2alm/beam_gauss.py#158)):
$b_\ell = \sqrt{\frac{2 \ell + 1}{4 \pi}}\exp{(-\frac{1}2 \sigma^2 \ell(\ell+1))}$

This is the reason of ell dependency. 
And reason of factor 2 is given by this [line](https://github.com/litebird/litebird_sim/blob/beam_comvolution_2/litebird_sim/beam_synthesis.py#L91) of lbs. 
It divides blm by 2, so lbs's blm is factor 2 smaller than grasp2alm.

I discussed the issue with Luca and we will address it with optics/beam experts during F2F.

paganol:
Hi @mreineck and @ziotom78, with @martamonelli we potentially found the bug of mueller_convolver. Simply inverting the angle of the HWP the output maps match. Here the comparison with scan map (assuming ideal HWP)
Screenshot 2025-01-20 at 18 39 07

mreineck:
If this has the blessing of @martamonelli, I'm totally happy with it!
@martamonelli, does this also resolve (some of) the remaining discrepancies with `beamconv`?

martamonelli:
Hi all,
the equations (and their derivations) shared by @okayamayu8 make perfect sense to me. Together with @paganol, we came up with a few tests (and minor changes to lbs) that will help us make sure the convolution is doing what it's supposed to. Let's see.

@mreineck: I started working on checking what changes for the comparison with beamconv, but I'm not done yet. I'll let you know as soon as I'm done.

mreineck:
Sorry for the long delay, @martamonelli, and thanks a lot also for te verification work! Having a tested Mueller convolver is a huge milestone!

litebird_sim/beam_convolution.py
mreineck: I think this should be an exception instead of a warning. What do you think, @paganol? If a pointing must be supplied and it isn't, there's no sense in continuing, correct?

martamonelli:
@mreineck, I've checked whether the changes proposed by Yuya to Mueller convolver take care of the discrepancies with beamconv and they don't. (This does make sense though: Yuya's formulae for the beam convolution assumes the same conventions as scan map, which seem to be different than the ones used in beamconv.)

mreineck:
Thank you for looking into this!
I'm glad it is a convention issue (although that doesn't make things any less complicated :-) and will port the changes back to my master version of Mueller convolver with additional comments to clarify this.

paganol:
Good news! I made some tests, after having merged master, and having adjusted the interface. The results are consistent with the previous implementation if the beams are rotated by the proper polarization angle. 
This simple piece of code, for a Gaussian beam, is sufficient:

```
def make_gauss_beam(fwhm_rad,lmax,mmax,psi_pol_rad):
    blm = lbs.gauss_beam_to_alm(lmax,
                                mmax,
                                fwhm_min_rad=fwhm_rad,
                                fwhm_max_rad=fwhm_rad,
                                psi_ell_rad=0,
                                psi_pol_rad=psi_pol_rad,
                                cross_polar_leakage=0)
    return blm

dets = []
blms = {}
for n_det in detlist:
        url=f"/releases/{imo_version}/satellite/{telescope}/{channel}/{n_det}/detector_info",
        imo=imo,
    )
    dets.append(det)
    blms[det.name] = make_gauss_beam(fwhm_rad,lmax,mmax,psi_pol_rad=det.pol_angle_rad)

```
btw.. potentially we could add a function like this to the code that automatically:
- reads fwhm, ellipticity, angle of the ellipses and polarization angle
- synthesize the elliptical gaussian beam for each detector
- store them in a dictionary. 

But we can discuss this later.. 

Regarding the comparison, here the results for Gaussian beams:
this with `scan_map` assuming ideal HWP 
Convolution

this for non ideal HWP,  assuming for each detector an ideal HWP + a random perturbation 

Convolution

I think we accomplished the first part of the validation. @martamonelli, @okayamayu8, @mreineck, @ziotom78, @patanch, I guess we can move to the asymmetric beam validation and also work a bit on the optimization.

Last thing.. the ~0.2% normalization is still there.. whatever I do.. any idea?

paganol:
Hi @mreineck, I also thought it was ringing at first, but I believe it's actually T-to-P leakage caused by the coupling of a non-ideal HWP and the scanning strategy. When I feed (0, Q, U) into the simulation, this is the result I get:

Convolution

Probably the ringing-like pattern is due to the scanning strategy..  does that sound convincing to you?

mreineck:
I did think of a purely scanning strategy related issue, i.e. that the equation system that needs to be solved for Q/U was ill-conditioned in some pixels. But that would have affected the ideal HWP as well, I guess. The leakage theory sounds much more convincing to me, thanks!

mreineck:
I still have absolutely no clue about the normalization issue; this almost feels as if there was a `pi=3.14` somewhere in the code ... or a division by `Npix-1` instead of `Npix`.

litebird_sim/beam_convolution.py
```diff
@@ -0,0 +1,359 @@
...
+        tod_det += fullconv.signal(ptg=pointings_det, alpha=hwp_angle)
+
+
+def add_convolved_sky(
+    tod,
+    pointings,
+    sky_alms: Union[SphericalHarmonics, Dict[str, SphericalHarmonics]],
+    beam_alms: Union[SphericalHarmonics, Dict[str, SphericalHarmonics]],
+    hwp_angle: Union[np.ndarray, None] = None,
+    mueller_hwp: Union[np.ndarray, None] = None,
```
ziotom78, line 142: Passing `mueller_hwp=None` make the code crash because of line #183:

```python
mueller_matrix=mueller_hwp[detector_idx],
```

litebird_sim/beam_convolution.py
```diff
@@ -0,0 +1,359 @@
...
+    """ """
+
+    n_detectors = tod.shape[0]
+
+    if type(pointings) is np.ndarray:
+        assert tod.shape == pointings.shape[0:2]
+
+    for detector_idx in range(n_detectors):
+        if type(pointings) is np.ndarray:
+            curr_pointings_det = pointings[detector_idx, :, :]
```
ziotom78, line 158: Unlike the `else:` branch, this does not initialize `hwp_angle`, which might be `None`

litebird_sim/beam_convolution.py
```diff
@@ -0,0 +1,359 @@
...
+        convolution_params = BeamConvolutionParameters(
+            lmax=default_lmax,
+            mmax=default_mmax,
+        )
+
+    else:
+        assert convolution_params.lmax - 4 >= convolution_params.mmax, (
+            "Error in the convolution parameters m_max must be ≦ ℓ_max!"
+            "Here ℓ_max={lmax} and m_max={mmax}!"
+        ).format(lmax=convolution_params.lmax, mmax=convolution_params.mmax)
```
ziotom78, line 82: The assertion error is inconsistent with the text.

litebird_sim/beam_convolution.py
paganol: This should be ok, if `hwp_angle ` is None `add_convolved_sky_to_one_detector` uses the standard convolution

litebird_sim/beam_convolution.py
paganol: Right

paganol:
Hi all, we're almost ready to merge! I added some documentation and updated the docstring. Now we need to decide if our current validation is sufficient and which tests we should implement. 
I've already included a couple of tests in the documentation that we can use for inspiration.

paganol:
@ziotom78, it seems that the decorator `_profile` conflicts with the possibility of accessing the docstring.
For example with the decorator:
```
Signature: sim.convolve_sky(*args, **kwargs)
Docstring: <no docstring>
File:      ~/Documents/Universita/litebird/simteam/codes/litebird_sim/litebird_sim/simulations.py
Type:      method
```
Without:
```
Signature:
sim.convolve_sky(
    sky_alms: Union[litebird_sim.spherical_harmonics.SphericalHarmonics, Dict[str, litebird_sim.spherical_harmonics.SphericalHarmonics]],
    beam_alms: Union[litebird_sim.spherical_harmonics.SphericalHarmonics, Dict[str, litebird_sim.spherical_harmonics.SphericalHarmonics]],
    input_sky_alms_in_galactic: bool = True,
    convolution_params: Optional[litebird_sim.beam_convolution.BeamConvolutionParameters] = None,
    component: str = 'tod',
    append_to_report: bool = True,
    nthreads: Optional[int] = None,
)
Docstring:
Fills the TODs, convolving a set of alms.

This method must be called after having set the scanning strategy, the
instrument, the list of detectors to simulate through calls to
:meth:`.set_instrument` and :meth:`.add_detector`, and the method
:meth:`.prepare_pointings`. alms are assumed to be produced by :class:`.Mbs`
File:      ~/Documents/Universita/litebird/simteam/codes/litebird_sim/litebird_sim/simulations.py
Type:      method
```
How can we fix it?

paganol:
Shall we proceed with the merge?

ziotom78:
> Shall we proceed with the merge?

Absolutely, go on!

ziotom78:
Be sure to update the file `STATUS.md`!

Issue #336, opened 2024-10-23
HWP angle


paganol:
The array stored in `observation.hwp_angle` (or returned by `observation.get_pointings()`) contains twice the hwp angle, this has a historical reason and it is related to the way the pointing was handled before. Should we reconsider this behavior, and return just the hwp angle? What do you think, @ziotom78 @sgiardie @nraffuzz @ggalloni? Btw, this is also relevant for the beam convolution module #170.
Also the documentation is not very clear on this aspect, and we should clarify this. 
```
@njit
def _get_ideal_hwp_angle(
    output_buffer, start_time_s, delta_time_s, start_angle_rad, ang_speed_radpsec
):
    for sample_idx in range(output_buffer.size):
        angle = (
            start_angle_rad
            + (start_time_s + delta_time_s * sample_idx) * 2 * ang_speed_radpsec
        ) % (2 * np.pi)

        output_buffer[sample_idx] = angle
```

PR #340, opened 2024-10-31
Redefinition of hwp_angle

paganol:
Here we address issue #336.


paganol:
This looks reasonable now, I've also updated the documentation that was still containing references to the old pointing API

paganol:
Hi @ziotom78 @ggalloni, I changed the pointing API as it was before this [commit](https://github.com/litebird/litebird_sim/pull/319/commits/b3bc3bb2049c152cc183d6cfc68f4598f5b93ec0). I've also improved the documentation. We should be almost ready to merge now. We just need to check the notebooks and update the CHANGELOG. 
Regarding a possible general reorganization of the documentation we can open a new PR for that.

Issue #341, opened 2024-11-06
Possible reorganization of the documentation


paganol:
The current version of the documentation is a bit messy, here are some suggestions we can implement to improve readability:
1. the scanning strategy chapter is rather long, we could move the hwp (and in general the pointing generation) in a new entry;
2. the chapter `Time order simulations` could be converted simple page with a general introduction and then links to single pages for dipole, scan map, noise, and (in the future) beam convolution.
3. all the systematics (non linearity, gain drift, hwp systematics, etc..) could go in a single separate chapter 
4. the entry Bandpasses could go in a sub entry of the chapter `Detectors, channels, and instruments`

anand-avinash:
If easily doable, I would suggest also to change the documentation theme. With the current theme I find it very difficult to navigate and infer the type, default values, etc. of the function arguments from the documentation. One such example:
![image](https://github.com/user-attachments/assets/2eb2fddd-ea2c-4664-812e-35c09a3a5f6a)

ziotom78:
Hi @anand-avinash , I agree, the look is terrible!

Do you have any example of a good alternative for our Sphinx theme? I have little experience in customizing Sphinx.

anand-avinash:
To me, PyData sphinx theme looks better. It is the same theme used by numpy and scipy docs
https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/gallery.html

yusuke-takase:
I tried to change a document of my [grasp2alm](https://yusuke-takase.github.io/grasp2alm/index.html#) `alabaster` (current lbsim's docs theme) to `pydata_sphinx_theme`, indeed it looks better and it was not too difficult.

ziotom78:
Hi all, thanks for your comments! I can work on this.

I am currently thinking of a possible way to get a reasonable outline for the manual. What do you think about the following structure?

![image](https://github.com/user-attachments/assets/87840afe-bb32-46e2-8f0a-69c2a8751205)

PR #342, opened 2024-11-13
Reorganize the User’s manual

ziotom78:
This PR addresses issue #341.

- [X] Use PyData as the default theme for docs and update Sphinx and deps
- [X] Rework the structure of the documentation
- [x] Check that the documentation is consistent and reads seamlessly


ziotom78:
I have completed fixing the grammar. Could somebody please have a look at the way the documentation looks now? Here is a screenshot with the new style:

image

paganol:
Hi @ziotom78, I like it, thanks!! I just added an example for the high level interface in the noise section. 

I also took the opportunity for changing the default behavior of `Simulation.add_noise`, before we had to pass explicitly the random number generator, now, if random is empty, `self.random` is used. This change doesn't really belong to this PR, so if you think that it should be in a separate PR I can do that, let me know.

https://github.com/litebird/litebird_sim/blob/bcf9cc07ef8977cea3069dc3903967a8ea7ff5f6/litebird_sim/simulations.py#L1549-L1574

ziotom78:
> I also took the opportunity for changing the default behavior of Simulation.add_noise, before we had to pass explicitly the random number generator, now, if random is empty, self.random is used. This change doesn't really belong to this PR, so if you think that it should be in a separate PR I can do that, let me know.

I know it’s boring, but having a separate PR for it would be clearer to users. This would let us properly advertise this change in the CHANGELOG by referencing the right PR.

paganol:
For me we can merge

Issue #189, opened 2022-08-29
Parallelize `generate_spin2ecl_quaternions`


ziotom78:
There is a long-standing `TODO` in [simulations.py](https://github.com/litebird/litebird_sim/blob/779618b2950ea31d621efd54c0ce8e4b8bc4a3c2/litebird_sim/simulations.py#L790-L795) that might affect the performance of simulations. It would be better to parallelize the call to `generate_spin2ecl_quaternions` when MPI is being used.

(See [this comment](https://github.com/litebird/litebird_sim/issues/184#issuecomment-1225910829) by @marcobortolami in issue #184.)

PR #344, opened 2024-11-14
Pointing systematics HWP class

yusuke-takase:
As I presented today's telecon, I have implemented the pointing systematics due to wedged transmissive HWP (or precessing reflective HWP) which causes a rotational disturbance to the pointings around the expected pointing direction. 
The HWP angle redefinition that happened recently (#340) is also covered. 

Current version of pointing-sys API has `Simulation` and `List[DetectorInfo]` as arguments like:
```python
pntsys = lbs.PointingSys(sim, dets_list)
```
However it was not useful for MPI distribution, so in this PR, the `Observation` is added in arguments as:
```python
pntsys = lbs.PointingSys(sim, obs, dets_list)
```
By this changing, we can multiply any systematic quaternions to any chunk of time given by MPI. 

And several methods that do quaternion multiplications were bit messy and redundant, so everything is unified to [this method](https://github.com/litebird/litebird_sim/blob/point_syst_wedgeHWP/litebird_sim/pointing_sys/pointing_sys.py#L140):
```python
lbs.left_multiply_syst_quats(
    result: RotQuaternion,
    syst_quats: RotQuaternion,
    detector: DetectorInfo,
    start_time,
    sampling_rate_hz
)
```

----------
Maybe following thing should not discussed here though let me point it out. 

An issue that I met in a preparation of [e2e-pipeline](https://github.com/yusuke-takase/e2e_pointing_sys/tree/master) for the pointing systematics is #189 .
Because this HWP-pointing systematics is happened in 19Hz intrinsically, so we forced to compute the `spin2ecliptic_quats`  in 19Hz as well to perform the `RotQuaternion` multiplication without stacking.

It requires a lot of memory because when we call the `Simulation.set_scanning_strategy()`, all MPI processes compute quaternions start to end of simulation.
It can be avoided by:
```python
(obs,) = sim.create_observations(
    detectors=dets,
    n_blocks_det=1,
    n_blocks_time=size,
    split_list_over_processes=False
)
scanning_strategy = lbs.SpinningScanningStrategy.from_imo(
    url= f"/releases/{imo_version}/satellite/scanning_parameters/",
    imo=imo,
)
sim.spin2ecliptic_quats = scanning_strategy.generate_spin2ecl_quaternions(
    start_time=obs.start_time,
    time_span_s=obs.n_samples/obs.sampling_rate_hz,
    delta_time_s=delta_time_s,
)
```
although if it is solved in clever way, it's definitely useful.


ziotom78:
Yeah, that is a long-standing TODO, but the idea I got at the time was to split the computation across the MPI processes and then gather all the quaternions in each process so that all the quaternions are available everywhere. This is, of course, not going to work in this specific case.

Your trick is probably the best way to do this. Perhaps we could add a paragraph or two about this in the documentation?

yusuke-takase:
Hi @ziotom78 ,

I wrote documentation of pointing systematics. I think it is ready to be merged into master if you are convinced.

ziotom78:
Thank you. I read the documentation, and it looks acceptable to me! Beware that you mistakenly added the `docs/build` folder to the repo. It is better to remove it, as ReadTheDocs automatically generates it, and it garbles the PR.

yusuke-takase:
Thank you for pointed it out, I didn't notice it. I removed it now.

PR #348, opened 2024-11-19
Fix bugs in default mueller set parameters definition

### LINKED ISSUE #335: minor bug in hwp_sys.set_parameters()
When using the [hwp_sys.py](https://github.com/litebird/litebird_sim/blob/master/litebird_sim/hwp_sys/hwp_sys.py) code using Mueller as the input matrix (in set_parameters) I think I found one small bug, because I get this error:

>   File "/home/mgomes/litebird/lbs_env/lib/python3.9/site-packages/litebird_sim/hwp_sys/hwp_sys.py", line 1071, in set_parameters
>     self.betas = np.deg2rad(self.betas)
> AttributeError: 'HwpSys' object has no attribute 'betas'

This is my set_parameters call:
```
hwp_sys.set_parameters(
    mueller_or_jones="mueller",
    Mbsparams=Mbsparams,
    Channel = lbs.FreqChannelInfo(bandcenter_ghz=140)
)
```

This is a code snippet of the region of **hwp_sys.py** code where the error comes from :

```
if mueller_or_jones == "jones":
    default_attrs = {
        "h1s": 0.0,
        "h2s": 0.0,
        "betas": 0.0,
        "z1s": 0.0,
        "z2s": 0.0,
    }

for attr, default_value in default_attrs.items():
    if not hasattr(self, attr):
        setattr(self, attr, default_value)
    self.betas = np.deg2rad(self.betas)
else:  # mueller_or_jones == "mueller":
    default_attrs = {
        "mIIs": 0.0,
        "mQIs": 0.0,
        "mUIs": 0.0,
        "mIQs": 0.0,
        "mIUs": 0.0,
        "mQQs": 0.0,
        "mUUs": 0.0,
        "mUQs": 0.0,
        "mQUs": 0.0,
    }

for attr, default_value in default_attrs.items():
    if not hasattr(self, attr):
        setattr(self, attr, default_value)
```

I would say that the _for_ block should be inside the if mueller_or_jones == "jones" block, and the last _for_ block should be inside the _else_ block (for the mueller case), because we don't have the beta parameter when using Mueller Matrix as input.

Also, I realized that, in this **hwp.sys** code, the jones matrix elements that are attributed by default when none are given correspond to the ideal case, while for the Mueller case they don't (they are all zero, instead of having the 1's in the diagonal). Is this intentional or is it something to change?

--

EDIT : I was going to create another issue, but since this error also comes from hwp_sys.set_parameters(), I add it in here. There might be another bug when giving a single frequency to the Channel parameter as in the example at the top.

>   File "/home/mgomes/litebird/lbs_env/lib/python3.9/site-packages/litebird_sim/hwp_sys/hwp_sys.py", line 984, in set_parameters
>     self.maps = mbs.run_all()[0][Channel.channel]
> KeyError: '140.0 GHz'

If, in line 984 of hwp.sys, I change [Channel.channel] to ['140.0_GHz'] it works, so the problem should be either in the definition of the dictionnary keys, or in the string coming from lbs.FreqChannelInfo(bandcenter_ghz=140), because the dictionnary does not accept '140.0 GHz' without the underscore as a key, however it is the one coming from the frequency channel definition. Or maybe I should not define the frequency channel like I did?

mj-gomes:
Closes #335. 

I also changed the multiple _"if not hasattr(self, attr):"_ to  _"if not paramdict.get(attr):"_ because, when assigning the value "False" to a non-given parameter, the _hasattr_ will consider the parameter already exists (is equal to False) and return **True**. 

This is only a problem when the argument "mueller_or_jones" == "mueller".  It must be because the parameters h1,h2,z1,z2,beta in the ideal case are all zeros and as False = 0, it works.

### DIFF
--- a/litebird_sim/hwp_sys/hwp_sys.py
+++ b/litebird_sim/hwp_sys/hwp_sys.py
@@ -959,24 +959,25 @@ class HwpSys:
                 }
 
                 for attr, default_value in default_attrs.items():
-                    if not hasattr(self, attr):
+                    if not paramdict.get(attr):
                         setattr(self, attr, default_value)
                 self.beta = np.deg2rad(self.beta)
+                
             else:  # mueller_or_jones == "mueller":
                 default_attrs = {
-                    "mII": 0.0,
+                    "mII": 1.0,
                     "mQI": 0.0,
                     "mUI": 0.0,
                     "mIQ": 0.0,
                     "mIU": 0.0,
-                    "mQQ": 0.0,
-                    "mUU": 0.0,
+                    "mQQ": 1.0,
+                    "mUU": -1.0,
                     "mUQ": 0.0,
                     "mQU": 0.0,
                 }
 
                 for attr, default_value in default_attrs.items():
-                    if not hasattr(self, attr):
+                    if not paramdict.get(attr):
                         setattr(self, attr, default_value)
 
             if np.any(maps) is None:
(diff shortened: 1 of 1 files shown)


ziotom78 (approved):
Great job, thanks!

ziotom78 (approved):

mj-gomes:
This pull request is no longer needed, as the new hwp_sys.py version to be merged does not have this issues anymore.

PR #358, opened 2025-02-03
Generalized optical chain

paganol:
This PR comes after some discussions we had with @martamonelli and @okayamayu8. 
It introduces some new features
1. The class `DetectorInfo` has three new args
- pol_angle_rad: polarization angle of the detector in radiants. The default is 0.
- pol_efficiency: polarization efficiency of the detector, defined as γ of eq. 15 of astro-ph/0606606. The default is 1.
- mueller_hwp: mueller matrix of the HWP for the detector. The default is None (i.e. no HWP)
2. The class `Hwp` has a new arg, i.e. mueller_hwp
3. `scan_map` now can handle:
- polarization efficiency for each detector
- separately a polarization angle of the detector and the orientation of the detector (telescope)
- both an ideal scanning (with an ideal hwp) and a generalized scanning with a generic hwp

If we implement these changes, the meaning of the quaternions in the imo will be different. Quaternions will not contain anymore the polarization angle but we keep separate the orientation of the detector and its polarization angle. 
These changes impact also PR #338. Once merged, we'll have a mueller_hwp matrix for `beam_convolution`, @mreineck @ziotom78.
@sgiardie, @nraffuzz, @mj-gomes, also the interface for `hwp_sys` can be simplified, having the polarization angle for each detector in the observations. 

The essential code is there, we still miss
- [x] to check if some optimization is possible
- [x] to fix the issue in `io.py`. It seems that it cannot handle a matrix per detector. Any idea here for an elegant solution, @ziotom78?
- [x] to update the map-makings (if necessary)
- [x] to update the documentation
- [x] update the mock IMo with updated quaternions


mreineck:
Thanks a lot for doing all this!

If the performance deficits occur in any area of the code I can help with, I'm very happy to do so. Unfortunately, conolution with a realistic beam and general HWP requires a lot of memory and CPU time, so this will never be a very quick process :-/

paganol:
Thanks @mreineck! I think it should be possible to optimize the numba implementation of the generic optical chain, in particular the function `scan_map_generic_hwp_for_one_detector`. I agree that convolution is probably already quite optimized, but this PR does not touch the convolution, it only deals with already convolved maps.
I see it more as a preparatory work. We introduce some quantities that will be useful for the convolution code (mueller matrix hwp and separation of orientation and polarization angle), and we provide potential validation opportunities for it.

paganol:
Thanks @ziotom78! The documentation may not be very detailed but it contains the main information. If you agree, I would merge asap, so we can move on to PR #338. Eventually we could review it when the convolution code is ready as well.

Compatibility with `hwp_sys` could also be broken, but we can deal with that in another PR.

ziotom78:
> We must modify this function and fix pointing systematics test and I can address the issue. The test loads the [mock_focalplane.toml](https://github.com/litebird/litebird_sim/blob/master/test/pointing_sys_reference/mock_focalplane.toml) first. (the toml file is generated by this [script](https://github.com/litebird/litebird_sim/blob/master/test/pointing_sys_reference/gen_mock_focalplane.py). ) I can re-genarate this toml file based on new `DetectorInfo` structure but I also think if new mock imo has multiple detectors with different quaternions, I can rewrite the test more coherent by using the imo without making toml file. Dose new mock imo have multiple detectors which have different quaternion?

Thanks, @yusuke-takase. It would be fantastic if you could fix `left_multiply_syst_quat()`!

Regarding the quaternions in the PTEP IMo, there is no variation: all of them are just `[0, 0, 0, 1]` (identity rotation). Putting more information in them would probably reveal details in the design of the focal plane that cannot go in a public repository…

mreineck:
I'm a bit confused, because adding a general Mueller matrix to the optical system will already introduce additional 2f and 4f harmonics into the generated TOD. Does the fact that a detector is looking at the HWP at an angle introduce yet more 2f and 4f contributions?

Just to be on the safe side: the Mueller matrix taken by `MuellerConvolver` corresponds to the "unrotated Mueller matrix" appearing as `M_HWP` in Eq. (A27) of https://arxiv.org/pdf/2012.10437.

mj-gomes:
@mreineck The problem is that, when we add the dependence on the angle of incidence, the 2f and 4f contributions are not the same. Eqs 2.5 - 2.7 in [https://arxiv.org/pdf/2308.00967](https://arxiv.org/pdf/2308.00967). The [M_0] elements would be the elements of the input mueller matrix, which needs in this case to be actually separated into 2 matrices, one for 2f and another for 4f.

@paganol Yes, it would be perfect if the DetectorInfo class has those two new arguments (or maybe the mueller_hwp argument can be Dictionary instead, in which each harmonic is a key).

paganol:
Hi @mj-gomes, I meant that we can add any number of new fields to the `DetectorInfo` class. Of course we need to discuss the proper object to add, and probably the IMo team can help here. Your idea of the dictionary seems good to me, we just need to check if this is compatible with the class `Observation`. 

All of this said, I would like to merge this PR asap, what about it we merge it as it is, so we can work on the PR #338, and we address the handling of the hwp mueller in another PR? 
Probably it will take a bit of time to converge, and it would be nice to create the most generic object possible that can accomodate many effects, and serve many modules.

mreineck:
@mj-gomes, thanks for the clarification!
If it turns out that we can save resources (mostly RAM, I guess) by tighter integration of your code with `MuellerConvolver`, I'm happy to make the necessary adjustments!

litebird_sim/detectors.py
```diff
@@ -132,6 +132,16 @@ class DetectorInfo:
              rotation at all, i.e., the detector is aligned with the
              boresight direction.
 
+        - pol_angle_rad (float): polarization angle of the detector in radiants. The default is 0.
```
ziotom78, line 135: I would explain this better: “polarization angle of the detector with respect to the x-axis of the reference frame of the detector”.

ziotom78 (approved):
This looks ok to me!

PR #363, opened 2025-02-19
HWP Systematics Non-Orthogonality Model Implementation

mj-gomes:
This pull_request implements [Patanchon et al. 2023](https://arxiv.org/abs/2308.00967)'s model into litebird_sim.

**set_parameters() changes:**

- the method no longer deals with the mueller/jones matrix elements, as these are stored in the DetectorInfo class. Hence:
the mueller_or_jones argument was removed. All the mueller elements (mII, mIQ, …) including ones for the mapmaking (mIIs, mIQs, …) were also removed.

- self.ata and self.atd initialization are now in set_parameters, as their initialization must be done outside of the observation loop.       

**fill_tod() changes:**

- xi angle comes now from pol_angle_rad argument form DetectorInfoClass.
- all mueller matrix elements are now obtained from the Detector objects.
- fill_tod() is now working with lbs parallelization (diving the observation into detector blocks as implemented by Avinash).

**TOD Computation Methods:**

- All functions related to the TOD computation were modified to include the rotation harmonics expansion.
- These functions are changed knowing that, because of the expansion in harmonics of the rotation frequency, the hwp rotation terms are now part of the Mueller Matrices elements, and hence the algebraic expressions were changed in that sense. Basically we went from this:

$$\boldsymbol{S'} = R(\psi_0) \ M_{HWP}  \ R(\rho-\psi) \  R(\psi) \ \boldsymbol{S}$$

- To this:

$$\boldsymbol{S'} = R(\psi_0) \ M_{HWP}(\Theta, \rho-\psi)  \ R(\psi) \  \boldsymbol{S}$$

where $\Theta$ is the incidence angle , $$\rho$$ is the half-wave plate rotation angle w.r.t the sky, $$\psi$$ is the spacecraft rotation w.r.t the sky, and $\psi_0$ is the polarization angle. S is the input sky, S' is the output one, and $M_{HWP}$ is the hwp matrix. The mueller matrix is given by the sum of three matrices, one for each of the harmonics (0f, 2f, 4f, where f is hwp's rotation frequency):

$$M_{HWP}(\Theta, \rho-\psi) = M_{ij}^{0f}(\Theta) + M_{ij}^{2f}cos(\Theta, 2\rho-2\psi) + M_{ij}^{4f}cos(\Theta, 4\rho-4\psi)$$

And the 2f and 4f terms are given by:

$$M_{ij}^{2f}(\Theta, 2\rho-2\psi) = M_{input_{ij}}^{2f}(\Theta) \  cos(2\rho - 2\psi - \phi _{ij}^{2f})$$

$$M_{ij}^{4f}(\Theta, 4\rho-4\psi) = M_{input_{ij}}^{4f}(\Theta) \  cos(4\rho - 4\psi - \phi _{ij}^{4f})$$

The $\phi_{ij}^{nf}$ values are simulated phases. The $M_{input}$ matrices are the constant parts dependent on $\Theta$, hence on the detector position, and **these are the matrices that need to be given as an argument of each detector**. Because $M_{ij}^{0f}(\Theta)$ elements do not have the $\rho-\psi$ dependency, we have that $M_{input}^{0f}(\Theta) = M^{0f}(\Theta)$.

**From a simulation point of view:**

- It is still possible to do everything that was done before, and even more: one can still for example introduce QI/UI only non-ideal terms to test for IP leakage, QU/UQ terms for QU mixing, and more.
- However, now it is possible to do this for each of the harmonics of the rotation frequency, as each of them has its own mueller matrix.
- The definition of an ideal HWP in the harmonics expansion is a bit less intuitive than usual. In fact, the mueller matrix for an ideal rotating hwp is:

$$M_{HWP_{ideal}} = \begin{pmatrix} 1 & 0 & 0\\\ 0 & cos(4\rho - 4\psi) & sin(4\rho- 4\psi) \\\ 0 & sin(4\rho- 4\psi) & -cos(4\rho- 4\psi) \end{pmatrix}$$

Which, in the hwp's reference frame ($$\rho-\psi=0$$), gives the usual matrix for an ideal hwp (diagonal 1, 1, -1). Because of the expressions for each mueller element shown above, we can see that, to obtain this ideal case, the input mueller matrices are:

$$M_{input}^{0f} = \begin{pmatrix} 1 & 0 & 0\\\ 0 & 0 & 0 \\\ 0 & 0 & 0 \end{pmatrix}; \ \ \ \ \ \  M_{input}^{2f} = \begin{pmatrix} 0 & 0 & 0\\\ 0 & 0 & 0 \\\ 0 & 0 & 0 \end{pmatrix}; \ \ \ \ \ \  M_{input}^{4f} = \begin{pmatrix} 0 & 0 & 0\\\ 0 & 1 & 1 \\\ 0 & 1 & 1 \end{pmatrix} $$

The sin and negative cosine terms are accounted for by the phases $$\phi$$ in the expressions shown above, which is (in an ideal case) close to $\pi$ for the 4f UU element, and to $$\frac{\pi}{2}$$ for IU and UQ (sin terms).

So, if one wants, for example, to see the IP leakage at 4f, one would give the input matrices as follows:

$$M_{input}^{0f} = \begin{pmatrix} 1 & 0 & 0\\\ 0 & 0 & 0 \\\ 0 & 0 & 0 \end{pmatrix}; \ \ \ \ \ \  M_{input}^{2f} = \begin{pmatrix} 0 & 0 & 0\\\ 0 & 0 & 0 \\\ 0 & 0 & 0 \end{pmatrix}; \ \ \ \ \ \ M_{input}^{4f} = \begin{pmatrix} 0 & 0 & 0\\\ 3 \times 10^{-4} & 1 & 1 \\\ 3 \times 10^{-4} & 1 & 1 \end{pmatrix} $$ 

- Finally, there is also a new method called mueller_interpolation(), which implements the interpolation equation which returns an estimate value for a given mueller element for a given incidence angle $\Theta$, while we don't have the simulated values.

**Essentially, it is done. Here's the missing points:**

- [ ] Band integration
- [ ] Method to go from given Jones input matrices to the different harmonics Mueller matrices (because it is not linear jones 4f -> mueller 4f, for example). Right now the code only works with Mueller matrices.
- [ ] There are some memory problems when paralelizing the fill_tod function, as the get_pointings() function uses, for example, about 500Gb when using 20 cpus at the same time.


mj-gomes:
I forgot to write the expressions including the $\Phi$ angle, that comes later in the article. It is the azimuthal angle of each detector's location in the focal plane, and is of particular importance in the multi-detector case (see equation 5.2 from [Patanchon et al. 2023](https://arxiv.org/abs/2308.00967)). This angle is also already included in the code's computations.

ziotom78:
Hi @mj-gomes , thanks for the work, and great job with your exhaustive comment!

I fixed the problem with poetry, it was already fixed in another PR and I backported the solution here. However, the tests do not pass because of a few problems:

1. Apart from a trivial conflict in `.github/workflows/tests.yml`, there is another conflict in `litebird_sim/hwp_sys/hwp_sys.py` that is related to the `xi` variable. I would leave this fix to you or somebody else who is better knowledgeable of the code.
2. The code should be reformatted using `ruff format`
3. Tests with `ruff check` do not pass because there are a few undeclared variables in the code.

mj-gomes:
I was trying to rewrite the test_hwp_sys.py code but in fact this test is not applicable to this new version: it uses reference values for the tod computed for an hwp band integration with a set of jones matrices, but the new model does not work for now with jones matrices. Also, I don't have reference values for an expected TOD. The best I can do for now is to test for the ideal hwp case to show it agrees with the expected. Should I do this?

mj-gomes:
> Hi @mj-gomes , not sure if this can help, but you can think to rewrite the current test using the Mueller matrix corresponding to the Jones parameters used. Then this should be (I guess) easy to compare to the new code, assuming that this Mueller matrix refers to the case with orthogonal incidence.

Hi @sgiardie, that is what I tried to do. The problem is that, when I convert from Jones to Mueller, I get a single matrix, and I need to have three. I thought about reverse engineering to get the three matrices, but it does not really work: in the new model, I need to develop the Mueller matrix into the harmonics of the rotation frequency. If I have a unique Mueller matrix, in order for it to be applicable to this model, I should be able to **represent in the harmonics expansion terms what is obtained by multiplying that unique matrix with the hwp rotation matrix**.

So, this is what we get applying to rotation:

$$\begin{pmatrix} mII & mIQ & mIU \\\ mQI & mQQ & mQU \\\ mUI & mUQ & mUU \end{pmatrix} \cdot \begin{pmatrix} 1 & 0 & 0 \\\ 0 & cos(4\rho-4\psi) & sin(4\rho-4\psi) \\\ 0 & -sin(4\rho-4\psi) & cos(4\rho-4\psi) \end{pmatrix} = \begin{pmatrix} mII & mIQ cos(4\rho-4\psi) - mIU sin(4\rho-4\psi) & mIU cos(4\rho-4\psi) + mIQ sin(4\rho-4\psi) \\\ mQI & mQQ cos(4\rho-4\psi) - mQU sin(4\rho-4\psi) & mQU cos(4\rho-4\psi) + mQQ sin(4\rho-4\psi) \\\ mUI & mUQ cos(4\rho-4\psi) - mUU sin(4\rho-4\psi) & mUU cos(4\rho-4\psi) + mUQ sin(4\rho-4\psi) \end{pmatrix} $$

So the majority of the terms get this "double dependence" on the rotation of the hwp. Basically, while in the model, each of the elements of the "final Mueller matrix", is a sum of the corresponding element on the 0f, 2f and 4f input matrices, each multiplied by a cos factor, to make this work, I would need to make each of the elements depend instead on more than one of the elements of the input matrix, and on more than one cos/sin factor, which would change the ideia of the code.

I might be overlooking something pretty straightforward, though. If that is the case, please tell me!

mj-gomes:
> Hi @mj-gomes , not sure if this can help, but you can think to rewrite the current test using the Mueller matrix corresponding to the Jones parameters used. Then this should be (I guess) easy to compare to the new code, assuming that this Mueller matrix refers to the case with orthogonal incidence.

Hi @sgiardie, that is what I tried to do. The problem is that, when I convert from Jones to Mueller, I get a single matrix, and I need to have three. I thought about reverse engineering to get the three matrices, but it does not really work: in the new model, I need to develop the Mueller matrix into the harmonics of the rotation frequency, and the rotation of the hwp enters in each of the matrix terms. If I have a unique input Mueller matrix, coming from Jones, in order for it to be applicable to this model, I should be able to **represent in the harmonics expansion terms what is obtained by multiplying that unique matrix with the hwp rotation matrix**.

So, this is what we get by applying the hwp rotation:

$$\begin{pmatrix} mII & mIQ & mIU \\\ mQI & mQQ & mQU \\\ mUI & mUQ & mUU \end{pmatrix} \cdot \begin{pmatrix} 1 & 0 & 0 \\\ 0 & cos(4\rho-4\psi) & sin(4\rho-4\psi) \\\ 0 & -sin(4\rho-4\psi) & cos(4\rho-4\psi) \end{pmatrix} = \begin{pmatrix} mII & mIQ cos(4\rho-4\psi) - mIU sin(4\rho-4\psi) & mIU cos(4\rho-4\psi) + mIQ sin(4\rho-4\psi) \\\ mQI & mQQ cos(4\rho-4\psi) - mQU sin(4\rho-4\psi) & mQU cos(4\rho-4\psi) + mQQ sin(4\rho-4\psi) \\\ mUI & mUQ cos(4\rho-4\psi) - mUU sin(4\rho-4\psi) & mUU cos(4\rho-4\psi) + mUQ sin(4\rho-4\psi) \end{pmatrix} $$

So the majority of the terms get this "double dependence" on the rotation of the hwp. Basically, while in the model, each of the elements of the "final Mueller matrix" (for each detector and for each sample), is a sum of the corresponding element on the 0f, 2f and 4f input matrices, each multiplied by a cos factor for that sample, now to make this work, I would need to make each of the elements depend instead on more than one of the elements of the input matrix, and on more than one cos/sin factor, which would change the ideia of the code.

I might be overlooking something pretty straightforward, though. If that is the case, please tell me!

sgiardie:
Hi @mj-gomes, I see. So, does that mean that we don't have an analytical way of tracing back Guillaume's formalism to the Jones/Mueller formalism in the orthogonal case? Which also means that the code would not be back compatible.
Or do we have it, but it is complicated to implement?

mj-gomes:
> Hi @mj-gomes, I see. So, does that mean that we don't have an analytical way of tracing back Guillaume's formalism to the Jones/Mueller formalism in the orthogonal case? Which also means that the code would not be back compatible. Or do we have it, but it is complicated to implement?

Hi @sgiardie, yes, for now we don't have a way to go from Jones to Mueller in this formalism, neither from Mueller to Jones, at least not that I am aware of.

sgiardie:
I see. So, you basically will get inputs for the HWP Mueller matrix at 0f, 2f and 4f and not care anymore about how to get from the Mueller matrix of the un-rotated HWP to the rotated ones, right? (sorry, we probably already discussed that at IPMU!)

Still, if it's not too messed up, it would be nice to have an analytical way to trace the un-rotated HWP to the harmonic components you use now. Like, is it too complicated to implement the equations you were writing above? Don't know if I am missing something fundamental here

mj-gomes:
Hello,

After some changes and some clarifications from Guillaume, this pull request is ready to be merged, if the reviewers agree.

This first version is **limited to single-frequency and to Mueller Matrices** as inputs. 

There is also a new example Jupyter Notebook (in the hwp_sys/examples) folder that shows how to perform simulations with this new code.

I am working on the next version, that will include:
- a new method to convert from Jones rotation harmonics matrices to Mueller rotation harmonics matrices (the method will be developed starting today)
- band integration (the method can be reused from the last version of the code, but we need to have matrices for several frequencies, so for now it is dependent on the previous point). Some things from the previous band integration were left in the code as I will reuse them in the next version, but they are useless in the current version.

There is also an ongoing effort to solve problems with the parallelization, as it is very hard to sustain the amount of memory needed when using MPI parallelization, and Numba parallelization results in (a small but eventually important) loss of precision-[see here](https://github.com/litebird/litebird_sim/discussions/366) (in the version in this pull request, njit is on but its parallelization is off). Other options are being studied (e.g. dask) in parallel to trying to fix the current problems.


The realistic matrix values that will come in the future, for several incidence angles and for several frequencies, will make it possible to use this code for all cases.

ziotom78 (approved):
I checked your changes. That’s a very good job, bravo! I have no objections to merging it.

mj-gomes:
Thanks @ziotom78.  
Just to be sure that the two reviewers are ok: @paganol, can I merge this?

Issue #364, opened 2025-02-21
Multiple issues when some MPI processes are left unused


anand-avinash:
If some MPI processes are left unused (for example, while using attribute based detector blocks), it now breaks the workflow at many places. I am listing here some of them that I found:

```
Traceback (most recent call last):
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/playground/lbsim_set_n_blocks/mpi_set_n_blocks_test2.py", line 117, in <module>
    sim.prepare_pointings()
  File "/mnt/Data/Projects/uniroma2/coding/dev_litebirdsim/litebird_sim/litebird_sim/simulations.py", line 250, in profile_wrapper
    result = function(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_litebirdsim/litebird_sim/litebird_sim/simulations.py", line 1312, in prepare_pointings
    prepare_pointings(
  File "/mnt/Data/Projects/uniroma2/coding/dev_litebirdsim/litebird_sim/litebird_sim/pointings_in_obs.py", line 46, in prepare_pointings
    assert all(m is None for m in cur_obs.mueller_hwp), (
                                  ^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

```
Traceback (most recent call last):
  File "/mnt/Data/Projects/uniroma2/coding/dev_cosmomap2/playground/lbsim_set_n_blocks/mpi_set_n_blocks_test2.py", line 167, in <module>
    binned_maps = sim.make_binned_map(nside)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_litebirdsim/litebird_sim/litebird_sim/simulations.py", line 250, in profile_wrapper
    result = function(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_litebirdsim/litebird_sim/litebird_sim/simulations.py", line 1726, in make_binned_map
    return make_binned_map(
           ^^^^^^^^^^^^^^^^
  File "/mnt/Data/Projects/uniroma2/coding/dev_litebirdsim/litebird_sim/litebird_sim/mapmaking/binner.py", line 330, in make_binned_map
    nobs_matrix = _build_nobs_matrix(
[...]

PR #365, opened 2025-02-21
Patches to fix issues that come up when some MPI processes are left unused

anand-avinash:
Reference #364


anand-avinash:
The first commit solves the first three errors listed in the issue. For the 4th error, the solutions don't seem trivial to me. It might be possible to just make it work, but it won't be the right way to do it. We should probably think of better ways to avoid this kind of issues that might appear with every breaking change in future.

PR #375, opened 2025-03-28
Fix issue #374

### LINKED ISSUE #374: Specifying pointings_dtype in scan_map_in_observations
Hi all,

I noticed something which is not really a bug but may eventually lead to problems.

When we call scan_map_in_observations, if we have not precomputed the pointings, it will compute them ([line 365](https://github.com/litebird/litebird_sim/blob/8baa320c8ec20c393cfa6b6a486c8e62336e20f9/litebird_sim/scan_map.py#L365)). However, it computes them without giving a value for pointings_dtype, which means it will compute them with the default single precision. 

I was running a simulation where I precomputed the pointings in double precision, and when I removed this precomputation, I noticed that the result maps were different. This shouldn't happen because precomputing the points should only be a matter of gaining time, and it should not yield a different result. Of course if we precompute the pointings in single precision, everything is ok.

Is there a reason for it to be like this? If not, my suggestion would be to add the pointings_dtype as an argument to scan_map_in_observations so that we can use it when calling get_pointings inside that method, so that we can be sure that we can control the precision independently of precomputing the points before the scanning or not.

mj-gomes:
Fixes #374.

It should be working correctly. Now we are able to define specifically pointings_dtype  when we call the following methods:

**in scan_map.py ->** scan_map_in_observations(), scan_map()
**in dipole.py ->** add_dipole_to_observations() and add_dipole()
**in mapmaking/destriper.py ->** make_destriped_map()
**in mapmaking/binner.py ->** make_binned_map()

I tried the scan_map_in_observations in the simulation in which I had noticed this issue and this change seems to be working. I actually never used the other methods so, for us to be sure, @anand-avinash if you can please try for a simulation that uses the mapmaking ones, and @paganol for the dipole ones, and confirm that there are no problems, it would be perfect. Thank you in advance!

### DIFF
--- a/docs/source/beam_convolution.rst
+++ b/docs/source/beam_convolution.rst
@@ -92,6 +92,7 @@ as demonstrated in the following example:
         blms,
         input_sky_alms_in_galactic=True,
         convolution_params=Convparams,
+        pointings_dtype=np.float32,
     )
 
     for i in range(obs.n_samples):
@@ -304,6 +305,7 @@ For a single-task execution, refer to the following example:
                      beam_alms=blms,
                      convolution_params=Convparams,
                      input_sky_alms_in_galactic=True,
+                     pointings_dtype=np.float32,
                      nthreads = 0)
 
 
--- a/litebird_sim/beam_convolution.py
+++ b/litebird_sim/beam_convolution.py
@@ -195,7 +195,7 @@ def add_convolved_sky(
     input_beam_names: Union[str, None] = None,
     convolution_params: Optional[BeamConvolutionParameters] = None,
     input_sky_alms_in_galactic: bool = True,
-    pointings_dtype=np.float32,
+    pointings_dtype=np.float64,
     nthreads: int = 0,
 ):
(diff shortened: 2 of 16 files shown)


mj-gomes:
I used single precision as the default because that's what was done before. If we decide to change to double, please tell me!

paganol:
Hi @mj-gomes, everything looks good to me. Thanks! 
I've added the same parameter to the beam convolution code and to the high level interface in the class `Simulation`.
Regarding your question, following what @mreineck said I would move to pointing in double.

anand-avinash:
Thanks @mj-gomes, everything looks good to me!

mj-gomes:
Thank you @paganol and @anand-avinash. I changed to default double precision and had to add the pointings_dtype specification also to the beam_convolution.rst file, because of the strict_typing, so that the tests pass.

I will merge this and close the issue. If someone notices some place that we missed in lbs where this should be applied, please reopen this issue.

PR #379, opened 2025-04-23
Update plot_fp.py for the new imo

mj-gomes:
Hi! 

This PR aims to update the plot_fp.py for the structure of the new imo, and the fact that there is no longer a choice between different telescopes. It also updates the fields of DetectorInfo shown next to the plot when selecting a detector.

I tested it locally and it seems to be working fine. The detectors positions correspond (simmetrically) to the center of the used wafers in the current configuration for option1M. Below you can compare the plot and the corresponding configuration given by the optics team.

image

image

### DIFF
--- a/litebird_sim/detectors.py
+++ b/litebird_sim/detectors.py
@@ -150,8 +150,9 @@ class DetectorInfo:
              for the mapmaking. It allows to have a non-ideal HWP in the solver.
              The default is None (i.e. no HWP)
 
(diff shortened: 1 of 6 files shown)


ziotom78:
Great, thank you for having updated this code!

mj-gomes:
Merged. The symmetry that we see is a bit strange, but it must be a problem with the quaternions in the IMo, we are looking into it right now. It should not affect the plot_fp code, so I am merging it.

PR #383, opened 2025-04-30
Downgrade ducc0 from 0.35.0 to 0.34.0

ziotom78:
This PR downgrades the dependency on ducc0 from 0.35.0 to 0.34.0, as I and @paganol have discovered that ducc 0.35.0 prevents MPI_AllReduce from working correctly on the CINECA Galileo100 machine when running a script to create a map from the TODs of 48 processes.


mreineck:
Two initial ideas to investigate this:
- try if the problem still occurs with `ducc` 0.37.1
- build the smallest possible reproducer, compile with full debugging symbols and run it under `valgrind`. I'm not sure how helpful this will be, however, since `numba` etc. could generate a lot of false positives.

mreineck:
In the meanime I'll run the entire `ducc` testsuite under `valgrind`, let's see if that turns up something.

ziotom78:
>     * try if the problem still occurs with `ducc` 0.37.1

I and @paganol already tested this, and the problem is still present in 0.37.1.

>     * build the smallest possible reproducer, compile with full debugging symbols and run it under `valgrind`. I'm not sure how helpful this will be, however, since `numba` etc. could generate a lot of false positives.

Compiling ducc with debug information @ CINECA is not trivial, as it complains that pybind11 is missing. Meanwhile, I am just running Valgrind on the official duck 0.35.0 package provided by PyPI, but I am not seeing warnings caused by `ducc0.*.so`.

mreineck:
If `ducc0` isn't causing `valgrind` messages, I really don't know how it can confuse the system in a way that MPI no longer works ... sorry, I'm at a complete loss here. :-(

mreineck:
Which parts of `ducc0` does the failing code use?
Grepping through the LBS sources, I mainly see usage of `ducc0.healpix.Healpix_Base`, `ducc0.totalconvolve`, some FFT calls and `PointingProvider`. If we could narrow down the problematic component, I can try to read the source diffs between 0.34 and 0.35.

mreineck:
Thanks for the information! I *think* this must be related to my changes to `PointingProvider`, which allow cyclic repetition of the provided pointings. I don't have very many test cases for that part of the code, so with a bit of luck it's in there.

mreineck:
I have been staring at the diffs for quite soe time now, and the most likely reason for the breakage seems that I switched the vectorization classes (`simd.h`) from something homegrown to something that is provisionally accepted for the C++ standard and available in both `gcc` and `clang`.
I can switch between the new and the old version via compiler flags, and I can try to provide a binary wheel. Still, direct compilation on Cineca would probably be beneficial, at least in the longer term.

Issue #387, opened 2025-05-13
Function for nullifying TODs


paganol:
We should add a small function in the simulation class for nullifying the TODs

PR #389, opened 2025-05-13
Nullifying functions in `Simulation`

paganol:
This PR addresses #387


litebird_sim/simulations.py
```diff
@@ -1167,6 +1167,38 @@ def describe_mpi_distribution(self) -> Optional[MpiDistributionDescr]:
...
+        AttributeError
+            If an observation does not have the specified component.
+        """
+        for i, cur_obs in enumerate(self.observations):
+            if not hasattr(cur_obs, component):
+                raise AttributeError(
+                    f"Observation {i} does not have attribute '{component}'"
+                )
+
+            tod = getattr(cur_obs, component)
```
ziotom78, line 1195: This can be rewritten more idiomatically as:

```python
try:
    tod = getattr(cur_obs, component)
except AttributeError:
    raise AttributeError(f"Observation {i} does not have attribute '{component}'")
```

because `getattr` already raises an `AttributeError` exception is `component` does not exist within `cur_obs`.

ziotom78:
Great! It would be nice to add an MPI test in `test_mpi.py` so that we are sure that the code works even when multiple processes are used.

PR #395, opened 2025-05-20
Save tod hwp sys

mj-gomes:
This is an intermediate PR (in the sense that the [this PR](https://github.com/litebird/litebird_sim/pull/386)) will have bigger changes to the hwp_sys.py code.

This one brings together everything needed for the simulations in the hackaton, including the non-linearity coupling with hwp_systematics, and the option to save the TOD as a .npy file to use with external mapmakers.


mj-gomes:
Just changed the hwp_sys.py code so that (only) compute_signal_for_one_detector has @njit(parallel=True), following the conclusion of the discussion [#366](https://github.com/litebird/litebird_sim/discussions/366#discussion-8012304).

mj-gomes:
Ready to merge!

ziotom78:
Wow, this has been a lot of work! @mj-gomes, @silviamich, can this be merged now?

test/test_nonlinearity.py
```diff
@@ -1,8 +1,38 @@
...
+
+    bytesobj = (str(input_str) + str(user_seed)).encode("utf-8")
+
+    hashobj = hashlib.md5()
+    hashobj.update(bytesobj)
+    digest = hashobj.digest()
+
+    return int.from_bytes(bytes=digest, byteorder="little")
+
+
```
ziotom78, line 35: This was already defined in `litebird_sim/non_linearity.py`. Is it a cut-and-paste error?

mj-gomes: Oh, I just seen this. I don't know, I will leave this answer to @silviamich.

ggalloni:
I was having a look at the seeding of the non-linear part. 

Being exactly equal to the gaindrift case, will this introduce a correlation between the two effects? Is this expected?

silviamich:
@ggalloni 
Right, it's true that both effects affect the detector, but in principle, I wouldn't expect this correlation. I didn't think about that. What do you suggest as an alternative seeding? We need something unique for each detector and realization.

mj-gomes:
@ggalloni I think using random seeds would be the best option to overcome that. Because even if we use a different seed for gaindrift than the one for non_linearity, there will still be a correlation, right? It may be high value - low value instead of high value - high value or something else, but using seed e.g. 12345 for gaindrift and 45337 for non_linearity at all times may result in a correlation. I don't know if this makes sense.

If this makes sense, then using a random generator for the seed would solve. However the generation of this "random" hash would have to be uncorrelated as well, or we would have the exact same problem. The generation of the seed should be done with a generator that does not take a seed as input. Some of them use the current date & time, and I suppose the two methods will never be called exactly at the same time. Even if the script is exactly the same, I suppose there are generators for which a difference of a milissecond can completely change the generated hash, so I suppose they should always have uncorrelated random seeds.

mj-gomes:
> Is this what is happening now? It is not, right? If so, OK, maybe the two things are actually uncorrelated, but then we lose reproducibility...

Oh sorry, I did not think about reproducibility. In that case my suggestion is indeed no good. If the generator defined within the simulation ensures this, it is for sure the best option.

silviamich:
> @silviamich @mj-gomes (@anand-avinash), could you remind me why we cannot simply use the generator defined within the simulation itself? If we always use that, we should ensure reproducibility while avoiding correlations, cause every time the generation is used, it should increase the seed somehow.

Because we need a different seed for each detector. If we use the simulation seed (which will be a different one for each realisation I guess) then we have the same value for each detector, which is not what we want to simulate

mreineck:
Hi all,

I'm just reading up on this, so I'm sorry if I'm mentioning something you have already discussed and dismissed...

Generally, if you want to generate many different independent sub-RNGs from a "master" RNG, I strongly recommend looking at `numpy.random.SeedSequence` and its functionality. This is a well-thought-out design and should be pretty mature by now. If possible, I'd prefer this over self-implemented alternatives.

anand-avinash:
I agree with @mreineck's comment as well. What we should now implement is a function (using numpy's SeedSequence) that is capable to produce different generators based following parameters:

1. MPI rank
2. detector property
3. systematic effect

This is to ensure that we can generate sequence of random numbers on each MPI process which is:

1. reproducible
2. independent of other generators
3. unique to a detector or a group of detectors (common-mode example) for a given systematic effect (to avoid possible correlation between two systematics, as pointed by @ggalloni)

In any case, we can discuss this more in the PR #414 and merge this PR if everyone agrees.

mj-gomes:
I was testing the MPI parallelization after all these changes and just confirmed, as I feared, that the code doesn't work for all cases (it works for n_blocks_det > 1 if n_blocks_time == 1, and for  n_blocks_time > 1 if there is only one detector). 

The fact that sim.detectors does not get separated into the ranks means that I have to insert new detector attributes into the Observation class (like we do with mueller_hwp or pol_angle_rad) so that I have everything that I need in its correspondent rank after creating the observation blocks.

I am fixing this and since we are in a hurry I hope to have it done by tomorrow.

mj-gomes:
I am getting an error because github is trying to perform the mpi test I just created using 1 process. I did what @anand-avinash suggested and included the command to use 4 cpus with openmpi and mpich but it is also doing it sequencially which in this case will not work.

Is there a way to avoid this in github test definitions or should I change the code so that if MPI is not enabled, it does not set n_blocks to avoid the error ?

ziotom78:
> I am getting an error because github is trying to perform the mpi test I just created using 1 process. I did what @anand-avinash suggested and included the command to use 4 cpus with openmpi and mpich but it is also doing it sequencially which in this case will not work.
> 
> Is there a way to avoid this in github test definitions or should I change the code so that if MPI is not enabled, it does not set n_blocks to avoid the error ?

Hi @mj-gomes, so far, we have implemented MPI tests following this philosophy:

- All the code and the tests should run with an arbitrary number of MPI processes. If there are explicit requirements, these should be checked within the test itself (for instance, this is the case for [`test_observation_tod_two_block_time`](https://github.com/litebird/litebird_sim/blob/9bf014069ca2e9941a108b84fcd3dbf9390234dc/test/test_mpi.py#L223)) instead of relying on the script in the YAML file, because it is clearer. If there is a constraint on the number of processes due to the way the test is designed, it is better to have this check as close as possible to the source code of the test itself.

- MPI tests are always run with one process and N processes, where the actual value of N has been changed several times in the past. Currently, we run it with only two processes to minimize resource consumption in our CI builds.

- New MPI tests should be added within the file `test_mpi.py` and listed in the `main` function at the bottom. We tried several solutions to use PyTest in the past, but none of them were robust enough.

Is your test truly tailored for N=4 processes, or could it be adapted to run with N=2? In the latter case, you could just put an `if` at the beginning of the test and exit if `mpi_size ≠ 2`, and then list this new test in the [`main` function](https://github.com/litebird/litebird_sim/blob/9bf014069ca2e9941a108b84fcd3dbf9390234dc/test/test_mpi.py#L681) of `test_mpi.py`, so that it is executed automatically without patching the YAML file.

mj-gomes:
> Is your test truly tailored for N=4 processes, or could it be adapted to run with N=2? In the latter case, you could just put an `if` at the beginning of the test and exit if `mpi_size ≠ 2`, and then list this new test in the [`main` function](https://github.com/litebird/litebird_sim/blob/9bf014069ca2e9941a108b84fcd3dbf9390234dc/test/test_mpi.py#L681) of `test_mpi.py`, so that it is executed automatically without patching the YAML file.

Hi @ziotom78, thank you for the explanation. This specific test was made to make sure that the previous bug when setting both n_blocks_det and n_blocks_time bigger than 1, which can not be obtained with N=2. However, if it is a problem I can just not insert this test in the repo, and add instead a test in test_mpi.py just for testing the MPI parallelization in hwp_sys.

ziotom78:
> Hi @ziotom78, thank you for the explanation. This specific test was made to make sure that we had no more the previous bug when setting both n_blocks_det and n_blocks_time bigger than 1, which can not be obtained with N=2. However, if it is a problem I can just not insert this test in the repo, and add instead a test in test_mpi.py just for testing the simpler MPI parallelization (either division in time or in det, not both of them at the sime time) in hwp_sys.

Oh, I see. No, then I think this deserves a separate call in the YAML file with N=4, and possibly a separate Python file containing the test. We could then use this new Python file to put all those tests for which N=2 is not enough. (Hopefully, this list of tests will always be very short!)

What do you all think? Does this sound reasonable?

ziotom78:
> Sounds good. But in that case I should still add a if comm.size < 4 with a single block to avoid the current error, right? There is no way to make the test run only for N=4?

Unfortunately, PyTest executes all the files that start with `test_*.py`. There should be a way to tell PyTest to skip some of these, but probably the safest way is to insert the check you suggest (`comm.size < 4`) at the beginning of your test; in this way, the code is clearer to read because the intent is stated explicitly.

mj-gomes:
There is apparently a new problem when github tries to mpirun with 4 cpus.

> 
> There are not enough slots available in the system to satisfy the 4
> slots that were requested by the application:
>   python3
> Either request fewer slots for your application, or make more slots
> available for use.
> A "slot" is the Open MPI term for an allocatable unit where we can
> launch a process.  The number of slots available are defined by the
> environment in which Open MPI processes are run:
>   1. Hostfile, via "slots=N" clauses (N defaults to number of
>      processor cores if not provided)
>   2. The --host command line parameter, via a ":N" suffix on the
>      hostname (N defaults to 1 if not provided)
>   3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
>   4. If none of a hostfile, the --host command line parameter, or an
>      RM is present, Open MPI defaults to the number of processor cores
> In all the above cases, if you want Open MPI to default to the number
> of hardware threads instead of the number of processor cores, use the
> --use-hwthread-cpus option.
> Alternatively, you can use the --oversubscribe option to ignore the
> number of available slots when deciding the number of processes to
> launch.

I don't know the best way to proceed. Is this worth the change?

mj-gomes:
Thank you @anand-avinash for the fix and everyone who worked on this PR!

I noticed small things in the docs of hwp_sys that I changed. After the tests for that are done, this is ready to merge if everyone agrees.

Issue #314, opened 2024-04-11
read_list_of_observations does not distribute all the information across mpi tasks


paganol:
I'm using `lbs.read_list_of_observations` to read a bunch of observations. Something like this:
```
components = ['tod_wn_1f_30mHz'] 
observations = lbs.read_list_of_observations(
    file_name_list = [path_tod+'LB_'+telescope+'_'+freq+'_obs_rank'+str(i).zfill(4)+'.hdf5' for i in list_of_tods_per_tasks],
    limit_mpi_rank = False,
    tod_fields = components,
    )
```
I can read the observations and everything seems ok. Then when I try to collapse the tods in maps, `lbs.make_bin_map` fails and the issue seems to be related to the fact that `obs.det_idx` is empty in ranks ≠ 0 and, as a consequence, `obs.n_detectors` fails. 

See here.
When I print:
```
print(rank,obs.det_idx,list_of_tods_per_tasks)
```
This is what I get:
```
1 None [2 3]
0 [ 0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20 21 22 23] [0 1]
```

paganol:
I tried to reproduce this issue with the current master, and it doesn't seem to be there anymore. I close it. Thanks @ziotom78!

PR #396, opened 2025-05-20
Implement fix for #314

ziotom78:
This PR investigates the cause for issue #314


ziotom78:
Bug #314 is no longer present, but since this PR fixed another bug in the code, we are merging it nevertheless.

PR #397, opened 2025-05-20
Center pointing

paganol:
This PR introduces an optional feature in the beam convolution code to align detector pointings to the centers of their corresponding HEALPix pixels. This adds a nice feature useful for debugging and test. The new option is called `nside_centering` and you can trigger the centering in this way:

```
sim.convolve_sky(sky_alms=alms,
                 beam_alms=blms,
                 convolution_params=Convparams,
                 input_sky_alms_in_galactic=True,
                 nside_centering=64,
                )
```
One possibility is to fold this option in the convolution_params. Let me know what you prefer. 

I've also introduced here some changes in `pointings.py`, in particular in `_get_pointings_array`. Now if `hwp_angle` is passed as a numpy array it supersedes the value retrieved from the observation. This helps in using non ideal HWPs in `scan_map`, `beam_convolution` or in the mapmakers. 

I've also added here a new function of the class `Observation` that triggers the computation of the hwp angle. Before we were forced to compute the entire pointing and select the second field.

Docstrings are there, we need just some documentation, and probably a test for the centered pointing. I'll work on that in the next days.


paganol:
I think we are ready for merging it

Issue #403, opened 2025-05-22
Failing MPI tests hang but do not report any error


ziotom78:
#401 has not solved the problem with MPI tests, because if there is a failure, GitHub does not realize it must stop.

PR #405, opened 2025-05-22
Improve the handling of MPI processes

ziotom78:
- **Move notebook in the proper place**
- **Use MPI_COMM_WORLD.Abort() when exception occur**
- **Add a “MPI” section in the output report**


ziotom78:
This is meant to be a fix for issue #403

PR #408, opened 2025-05-23
Improve get_sky

paganol:
This introduces a new option in get_sky for obtaining maps/alms per channel instead of per detector.

### DIFF
--- a/docs/source/simulations.rst
+++ b/docs/source/simulations.rst
@@ -537,7 +537,7 @@ TODs of the simulations can be also zeroed easly with::
 
 See the documentation in :ref:`observations`, :ref:`scanning-strategy` 
 :ref:`dipole-anisotropy`, :ref:`noise`, :ref:`mapscanning`, :ref:`mapmaking`,
-:ref:`beamconvolution` for details of the single functions.
+:ref:`beamconvolution`, :ref:`Mbs`, for details of the single functions.
 
 
 Data splits
--- a/docs/source/sky_maps.rst
+++ b/docs/source/sky_maps.rst
@@ -64,6 +64,41 @@ multipole of these alms with ``lmax_alms``, the default value is
 the umbeamed maps or alms.
 
 
+Interface in the Simulation class
+---------------------------------
+
+The `Simulation` class provides a high-level method,
+:meth:`litebird_sim.simulation.Simulation.get_sky`, for quickly generating
+synthetic sky maps using Mbs parameters. This method internally constructs
+an :class:`.Mbs` instance and runs the simulation either based on the
+detectors involved in the current observations or using user-specified
+frequency channels.
+
+This interface is useful when synthetic maps are needed without direct
+interaction with the `Mbs` class. Here's a basic example:
+
+.. code-block:: python
+
+    import litebird_sim as lbs
+
+    sim = lbs.Simulation(base_path="../output", random_seed=42)
+    params = lbs.MbsParameters(
+        make_cmb=True,
+        make_fg=True,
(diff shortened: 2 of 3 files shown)


ziotom78 (commented):
So far, so good! What is missing is a simple test that checks that everything works fine when you pass a frequency channel.

ggalloni:
Now that the `alms` of the sky and `blms` of the beam can be stored in the observation, they are both optional in `sim.fill_tods` and `sim.convolve_sky`. So, I added a try-except that tries to retrieve them from the observation in both those routines.

I added a couple of tests:

1. If there is more than one detector from the same channel, I check that the maps produced for each detector when `get_sky` is used, as in the old implementation, are equal to the single map produced when I specify the channel to `get_sky`.
2. Since now the `alms` of the sky and `blms` are optional in `sim.fill_tods` and `sim.convolve_sky`, I test those routines without arguments. The ellipticities are all zeros in this case, giving me `nan` in the Gaussian beam. So, I must set them manually to one to make this work. Does this make sense to you, @paganol @ziotom78? You can test this by commenting line 752 of test_simulations.py and printing the blms.

paganol:
Hi @ggalloni, thanks!
What about if we put that check in `lbs.scan_map_in_observations` and `lbs.add_convolved_sky_to_observations` instead of in `Simulation.fill_tod` or `Simulation.convolve_sky`?

So, for example, this:

```
        if sky_alms is None:
            try:
                sky_alms = curr_obs.sky
            except AttributeError:
                msg = "'sky_alms' is None and nothing is found in the observation. You should either pass the spherical harmonics, or store them in the observations if 'mbs' is used."
                raise AttributeError(msg)
            assert sky_alms["type"] == "alms", (
                "'sky_alms' should be of type 'alms'. Use 'store_alms' of 'MbsParameters' to make it so."
            )

        if beam_alms is None:
            try:
                beam_alms = curr_obs.blms
            except AttributeError:
                msg = "'beam_alms' is None and nothing is found in the observation. You should either pass the spherical harmonics of the beam, or store them in the observations if 'get_gauss_beam_alms' is used."
                raise AttributeError(msg)
```
in  
```
def add_convolved_sky_to_observations(
    observations: Union[Observation, List[Observation]],
    sky_alms: Optional[
            Union[SphericalHarmonics, Dict[str, SphericalHarmonics], None]
     ] = None,
     beam_alms: Optional[
            Union[SphericalHarmonics, Dict[str, SphericalHarmonics], None]
     ] = None,
    pointings: Union[npt.ArrayLike, List[npt.ArrayLike], None] = None,
    ....
```
Similar thing can be done in `lbs.scan_map_in_observations`

Regarding the point 2 of your message, it's strange that ellipticities are all zeros, they should be initialized to 1 in the `DetectorInfo` class. Could you point me to the issue?

ggalloni:
I opened issue #411.

Anyway, for moving the checks, we have a choice to make, tho. Either we keep the check on `maps` and `sky_alms` where they are, since those are needed for the construction of the reports, or we make 'scan_map_in_observations' and `add_convolved_sky_to_observations` return them. What do you think?

paganol:
> Anyway, for moving the checks, we have a choice to make, tho. Either we keep the check on `maps` and `sky_alms` where they are, since those are needed for the construction of the reports, or we make 'scan_map_in_observations' and `add_convolved_sky_to_observations` return them. What do you think?

Right.. good point, maybe we could include the reports in `Simulation.get_sky`. I don't like introducing reports in `scan_map_in_observations ` or `add_convolved_sky_to_observations ` since they really belongs the simulation class.

ggalloni (approved):
In my opinion this is good to go.

Issue #412, opened 2025-05-25
Mbs saving maps in Galactic coordinates even if maps_in_ecliptic == True


martamonelli:
Not sure if it's an actual issue or just a feature, but I was assuming the code to be doing something else and just made a mess eheh

If I understand correctly, the rotation from Galactic to Ecliptic is only performed after the maps are saved. Therefore, if I set `save = True`, those maps will be saved in Galactic coordinates, regardless of whether `maps_in_ecliptic` is `True` or `False`.

paganol:
> Not sure if it's an actual issue or just a feature, but I was assuming the code to be doing something else and just made a mess eheh
> 
> If I understand correctly, the rotation from Galactic to Ecliptic is only performed after the maps are saved. Therefore, if I set `save = True`, those maps will be saved in Galactic coordinates, regardless of whether `maps_in_ecliptic` is `True` or `False`.

Correct, and this behavior was only partially intentional... the option for outputting ecliptic coordinates was added later, this is another thing we should check when we'll revise mbs

PR #413, opened 2025-05-25
Move rotation of maps at generation

ggalloni:
In response to #412, this PR moves the rotation of maps in MBS at generation, so that both saved and returned maps are rotated.

### DIFF
--- a/litebird_sim/mbs/mbs.py
+++ b/litebird_sim/mbs/mbs.py
@@ -335,6 +335,10 @@ class Mbs:
         self.ch_list = channel_list
         self.pysm_units = u.Unit(self.params.units)
 
+        self.rotator = None
+        if self.params.maps_in_ecliptic:
+            self.rotator = hp.Rotator(coord=["G", "E"])
+
     def _parse_instrument_from_det_list(self):
         self.instrument = {}
         try:
@@ -515,6 +519,10 @@ class Mbs:
                             + f"_{nmc_str}_{file_str}.fits"
                         )
                         cur_map_path = nmc_output_directory / file_name
+                        if self.rotator is not None:
+                            noise_map_split = self.rotator.rotate_map_alms(
+                                noise_map_split, lmax=self.params.lmax_alms
+                            )
                         lbs.write_healpix_map_to_file(
                             cur_map_path, noise_map_split, column_units=col_units
                         )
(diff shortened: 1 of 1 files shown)


martamonelli (approved):
Everything looks good to me

paganol:
Hi @mj-gomes, could it be caused by the PR #408? We changed the handling of the lmax in the CMB alm generation, see here 
https://github.com/litebird/litebird_sim/pull/408/commits/6b79eeef0e216fa6983ffd29405af037adf002bb
Now lmax_alms is used in synfast, before 3*nside-1 was always used.

mj-gomes:
Hi @paganol, I tried running a test without the change in that commit, and it does not look like that is what's changing.

mj-gomes:
Here are the differences (in Kelvin) for a case with cmb + solar dipole + fg (s0+d0+ff1)

image

mj-gomes:
> Hi @mj-gomes, if you use only one field, e.g. the foreground, do you see the same difference?

The difference comes from the CMB (see figure below). There is a small discrepancy also in fg but the order of magnitude is very small. It would be logical that the problem came from the synfast, but I tried again and it doesn't.

image

ziotom78:
These results are really weird! To recap:

- **Before** this PR, we rotated the map just once we summed all the contributions
- This PR rotates each map (dipole, CMB, …) immediately after its creation so that the maps are summed when they are already in the ecliptic reference frame

The difference is mostly due to CMB (10^-6 K), because FG are negligible (10^-20) and the dipole is zero.

Luca and I have taken a look at this but have no clues. @mreineck, do you have any idea what might be causing this?

mj-gomes:
> This might be related to the order of smoothing and rotation.
> 
> For the CMB in the old code we had:
> 
>     * synfast
> 
>     * smoothing
> 
>     * rotation
> 
> 
> In the new code we have:
> 
>     * synfast
> 
>     * rotation
> 
>     * smoothing
> 
> 
> This does not affect the foreground because the order is always: generation, smoothing, rotation. Does it sound reasonable?

This is it! I just tried inverting the order between rotating and smoothing for the CMB and it solves it!

Which order should we use? Is there an argument for preferring one of them over the other?

mreineck:
I'd also go for the first option.
Rotation of maps is always bad numerically, and should be avoided whenever possible. But if the maps are smoothed, it's a little less bad :-)

mreineck:
On the other hand, if we are talking about the CMB here, why not do `syn_alm` -> `rotate_alm` -> `alm2map`? That is prectically lossless.

paganol:
I totally agree, I was considering that solution as well.

PR #416, opened 2025-05-31
Fix the definition of ellipticity and fwhm

paganol:
This PR is meant to correct the definition of the ellipticity and fwhm used in the code.
So far we used, coherently with Planck [LevelS](https://github.com/zonca/planck-levelS/blob/master/Beam/gaussbeampol_main.f90)
- fwhm = sqrt(fwhm_max*fwhm_min)
- e = fwhm_max/fwhm_min

We also take the opportunity here to include in the documentation a section for the functions in `beam_synthesis.py`.


paganol:
We agreed to adopt the Planck/LevelS convention for defining ellipticity. I’m closing this PR. The upcoming PR #419 will update the ellipticity from 0 to 1 in the IMO vPTEP to ensure consistency with this convention.

Issue #417, opened 2025-06-04
no interpolation in hwp_sys.fill_tod


martamonelli:
I had a quick exchange with @mj-gomes and realized that `hwp_sys.fill_tod` doesn't take the `interpolation` parameter, but simply implements the default `scan_map` behaviour, i.e. the nearest-neighbor interpolation (see [here](https://github.com/litebird/litebird_sim/blob/9bf014069ca2e9941a108b84fcd3dbf9390234dc/litebird_sim/scan_map.py#L228)).

Shouldn't we iclude the option to have linear interpolation there as well?

mreineck:
> Just to confirm something: @mreineck, is ducc0's Healpix_Base just an interface for healpix?

Technically no. `ducc0` comes with its own implementation of many of the Healpix functions. So if you use the `Healpix_Base` class, you are using `ducc0` exclusively and have no dependency on `healpy`.
But they do the same thing :)

BTW: if I'm not replying to any of the messages I'm tagged in, it's because I made the mistake of fully subscribing to the `litebird_sim` repo ... and now I'm getting hundreds of emails per day, which I definitely cannot read ...
I have unsubscribed now and hopefully will now get only mails when I'm explicitly mentioned.

PR #420, opened 2025-06-05
Add linear interpolation to hwp_sys.py

mj-gomes:
This PR aims to answer the issue #417 and insert the possibility of using linear interpolation on the input_maps, in the same way as scan_map does.

### DIFF
--- a/litebird_sim/hwp_sys/hwp_sys.py
+++ b/litebird_sim/hwp_sys/hwp_sys.py
@@ -272,7 +272,6 @@ def compute_signal_for_one_sample(
 @njit(parallel=True)
 def compute_signal_for_one_detector(
     tod_det,
-    pixel_ind,
     m0f,
     m2f,
     m4f,
@@ -296,9 +295,9 @@ def compute_signal_for_one_detector(
         Four_rho_phi = 4 * (rho[i] - phi)
         Two_rho_phi = 2 * (rho[i] - phi)
         tod_det[i] += compute_signal_for_one_sample(
-            T=maps[0, pixel_ind[i]],
-            Q=maps[1, pixel_ind[i]],
-            U=maps[2, pixel_ind[i]],
+            T=maps[0][i],
+            Q=maps[1][i],
+            U=maps[2][i],
             mII=m0f[0, 0]
             + m2f[0, 0] * np.cos(Two_rho_phi - 2.32)
             + m4f[0, 0] * np.cos(Four_rho_phi - 0.84),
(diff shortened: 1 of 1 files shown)


mj-gomes:
@ziotom78 is this red coverage report a problem?

mj-gomes:
> The problem is that the new interpolation was not tested. I added a new test and removed a warning produced by Numba. (The latter change might also improve a bit the performance, but I do not expect huge savings.)

Thank you!

> BTW, I looked at the code and noticed that `HwpSys.set_parameters()` accepts two parameters (`Channel`, `Mbsparams`) with a capital letter, which is in contrast with our usual naming convention. Is there a strong reason for this? Otherwise, I would change them to small case.

There is no particular reason, I just left them as they were when I picked the code. I will change that.

> Also, I see that there is an `example` folder in `litebird_sim/hwp_sys` containing a Jupyter notebook and a script. Wouldn’t be better to move that notebook in the `notebooks` folder to enhance its discoverability? Regarding the script, I'm unsure where it could be placed.

Sure, the notebook can go to the notebooks folder. Regarding the script, I was thinking about it and perhaps it can be removed actually. It is specific to hwp systematics coupled with non linearity, which is something only Silvia and I do, it is maybe unnecessary to have it in the repo...

ziotom78:
> There is no particular reason, I just left them as they were when I picked the code. I will change that.

Ah, ok. I can do it, as my PyCharm is already opened, and I can quickly rename all the instances without introducing grammar errors. I’ll do that in another PR.

Once we get the green light from @martamonelli, this PR can be merged.

ziotom78:
Mmm… On second thought, I believe it is crucial to mark release 0.15.0 as soon as possible so that people can start the debug run. I will merge this PR, add the new PR with the name change (`Channel` → `channel`), and then I will release 0.15.0.

I am pretty confident that this PR is ok because the test @mj-gomes added is quite comprehensive. If there is some issue, we can release 0.15.1 later.

mj-gomes:
> Mmm… On second thought, I believe it is crucial to mark release 0.15.0 as soon as possible so that people can start the debug run. I will merge this PR, add the new PR with the name change (`Channel` → `channel`), and then I will release 0.15.0.
> 
> I am pretty confident that this PR is ok because the test @mj-gomes added is quite comprehensive. If there is some issue, we can release 0.15.1 later.

Yes, Marta will not give any news today I would say, given the Japanese timezone. I talked with her this morning and she applied the non-linearity and it changes things for the ideal hwp case, but the spikes are still there for the 2f case. I don't think this is a problem from the interpolation itself, it is pretty straightforward and basically the same thing as in scan map. So I would agree it is safe to merge.

PR #425, opened 2025-06-09
Improve MBS module

paganol:
Fixing the issue discussed in PR #413 is not trivial if we want to retain the current structure, especially if we intend to use [ducc0](https://gitlab.mpcdf.mpg.de/mtr/ducc/-/tree/ducc0). 
Rather than implementing a complex patch, I propose we use this as an opportunity to rework and simplify the MBS module.
The current MBS implementation is quite old, overly complex, and includes features that are no longer necessary for lbs.  

So far, I’ve only included some preliminary changes, i.e. include in `coordinates.py` the computation of the Euler angles that can be used in [ducc0's rotate_alm](https://mtr.pages.mpcdf.de/ducc/sht.html#ducc0.sht.rotate_alm). 
Next minimum goals for this PR should be:
* remove the noise generation, it's something we never used and it only creates confusion
* remove the mpi interface, it conflicts with the general mpi structure of lbs
* avoid many harmonic transform when it is not necessary
* remove the option for saving the maps, and provide a method in the Mbs class for exporting maps if needed.

In general `mbs.py` should only serve as input generator for lbs. For simplified generation of CMB, foregrounds, and white noise maps, we could build a new dedicated module outside lbs.

Plans for this PR:
- [ ] New functions in `coordinates.py `
- [ ] Reorganize mbs.py, potentially changing the interface
- [ ] Update tests 
- [ ] Make the rest of the code compatible
- [ ] Update documentation

@ziotom78, @NicolettaK what do you think?


ziotom78:
Hi @paganol, I agree with the approach!

paganol:
The development for this PR is now addressed in PR #479. We can close it.

PR #427, opened 2025-06-09
`MPI_COMM_GRID` cleanup

anand-avinash:
This PR is meant to carry forward what we decided in #372. Specifically, we will stop fixing the issue #364, and instead, provide a script for user to help with determining the number of MPI processes needed to run the simulation. In case a user provides more MPI slots than needed, the code will produce an error and direct the user to relevant info on selecting the correct number of MPI slots. With these changes, there will be no need for `MPI_COMM_GRID` as `MPI_COMM_GRID.COMM_OBS_GRID` will always point to `MPI.COMM_WORLD`. So in this PR we will:

- [x] include selective incremental changes from #372
- [ ] remove `MPI_COMM_GRID` and update the code accordingly

The script for determining the correct number of MPI slots will be implemented in a separate PR.

### DIFF
--- a/litebird_sim/__init__.py
+++ b/litebird_sim/__init__.py
@@ -82,7 +82,7 @@ from .io import (
 )
 from .madam import save_simulation_for_madam
 from .mbs.mbs import Mbs, MbsParameters, MbsSavedMapInfo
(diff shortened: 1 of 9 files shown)


anand-avinash:
Superseded by #539

PR #429, opened 2025-06-12
Fix pointings bug in hwp_sys.py

mj-gomes:
This PR fixes a bug in hwp_sys.py that made pointings recompute at single precision, even when one had precomputed them explicitly in double precision.

Now if the pointings are precomputed, whatever the precision, they will, as expected, automatically be used in hwp_sys and not computed again.


ziotom78:
Great work! This PR is particularly important, as it addresses a potential issue that could impact the KDP2 simulations.

Once we merge this, we should release LBS 0.15.1 immediately.

mj-gomes:
> Once we merge this, we should release LBS 0.15.1 immediately.

I would just suggest waiting for a PR that @martamonelli will open tomorrow that will also change some things in the on-the-fly mapmaker in hwp_sys. It is less important but since we are releasing a patch, it could perhaps also include this change.

Issue #424, opened 2025-06-06
Numerical errors in hwp_angle calculation


mj-gomes:
Hi,

I was trying to figure out Marta's 2f problem and realized that, when we compute the hwp_angle (hereafter called alpha), the % (2 * np.pi) operation is creating numerical errors. This is the method:

https://github.com/litebird/litebird_sim/blob/deaa35c0879e30338a696cf4dfa284f7a78b2d09/litebird_sim/hwp.py#L135-L144

Basically, there are some differences due to floating point arithmetics between cos(alpha) and cos(% (2 * np.pi)).

This is usually negligible but for a high value for the angle alpha, this difference increases. And when we go to e.g. alpha=1e8, which is the case for our simulations since the hwp_angle is proportional to the sample number, the difference between the two cosines goes up to 1e-9. And for the rotating hwp, we have cos(4*alpha) so this becomes even bigger.

Also, if we plot the evolution of this difference between the both cosines at 4alpha for alpha = (1e8 + x), where x varies between 0 and 2pi, so basically plotting how this cosine difference evolves over a full rotation of the hwp for a given time around about 6 months into the observation, we actually get a sinusoidal behaviour.

![Image](https://github.com/user-attachments/assets/18cff591-5573-4773-885c-957355e563de)

I don't know if this explains Marta's problem or not (I did asked her to try it just to be sure). I am a bit confused but talked with Eric about this and his intuition is that it probably won't. However, this is still intriguing because it means we are getting wrong values for our terms with cosine/sine(hwp_angle).

Is this something expected? Do we really need the modulus operator? I did not find any information in numpy's docs stating a max value for the angle in their trigonometric functions, so could we just use the real angle computed without performing the modulus?

ziotom78:
Hi @mj-gomes, thanks for the analysis and the plots! I do not believe that we can achieve a “perfect” solution here. The problem is that we are using finite-precision floating-point numbers together with a way to measure angles that relies on an infinite-precision number (π).

mj-gomes:
@ziotom78 I think I found a way to keep the precision. If we use the the [mpmath library](https://mpmath.org/), running the exact same code that gave me the previous plot, I get **perfect** precision with or without the modulus (plot below)

All I had to do was doing mpmath.cos and mpmath.pi instead of using numpy. Should we implement this in lbs?

![Image](https://github.com/user-attachments/assets/da12769b-eee0-472e-8d77-2fc199166874)

mj-gomes:
[...]

PR #433, opened 2025-06-18
Change angle computations to the complex domain

mj-gomes:
I am opening this PR following the discussion in [424](https://github.com/litebird/litebird_sim/issues/424), so that we can work on the changes mentioned.

To begin with, I have updated the hwp angle computation and edited the reference values in a test that otherwise would result in an assertion error. The angles are in fact the same as before (this test is done for some seconds so the loss of precision is not yet visible). However, while before the angles kept on increasing, in this new case the maximum angle is pi, and everything bigger than pi is instead a negative number. I confirmed that these angles where in fact the same as before.

I think now there are two things left to do, which are:

- [ ] change this for other angles being computed in the incremental way which results in a loss of precision
- [ ] refurbish lbs in several places so that we can use the cosines and sines directly obtained from the complex computations, in order to gain time (this would mean mix other angles in the calculus, for example the phases in hwp_sys computation).

EDIT: I added a normalization step, as it highly increases the precision (see https://github.com/litebird/litebird_sim/issues/424#issuecomment-2984725174).


mj-gomes:
Ready to merge! Getting cosine terms directly from the complex calculus will be done at a later stage.

PR #441, opened 2025-06-24
Advanced operators for `BrahMap`

ggalloni:
@anand-avinash I prepared this PR to start using more advanced operators, but tests are failing.

This is not urgent at all; it just serves as a place to add support for the new operators.


ggalloni:
@anand-avinash I noticed that the test on circulant fails if I make the duration longer than 70 seconds, and you have something similar in `BrahMap` tests. What is going on there?

PR #442, opened 2025-06-24
Set Mueller Matrix phases as attributes

mj-gomes:
This PR fixes an issue raised by Guillaume and Marta regarding the fact that the phases in the mueller matrices were hardcoded. This was a temporary measure but it's not ideal. I was asked to change it so that someone doing simulations is able to change the phases to adapt to the configuration.

The phases are now given by an argument of hwp_sys.set_parameters(), and in case nothing is given, the default phases are the ones that were hardcoded before, i.e., the ones in Table 1 in [Patanchon et al. 2021](https://arxiv.org/pdf/2308.00967).

Just add mueller_phases in the call to set_parameters() as a dictionary containing the matrix for 2f and 4f, like the one in:

https://github.com/litebird/litebird_sim/blob/58bb6c2edcafd353b31bca4dc50c0e30556f5f5c/litebird_sim/hwp_sys/hwp_sys.py#L613-L626


martamonelli (approved):
Looks good to me!

Issue #446, opened 2025-07-02
Algebra for SphericalHarmonics


paganol:
It would be useful to define algebraic operations for the `SphericalHarmonics` class to facilitate analytical manipulations and simulations. In particular, the following features would be valuable:

- Sum of multiple `SphericalHarmonics` objects
- Multiplication by a scalar $A$, or by a $A^i$ vector with $i\in{T,E,B}$
- Convolution by a $f_\ell$, or by a $f_\ell^i$ again with $i\in{T,E,B}$

PR #448, opened 2025-07-02
Algebra for SphericalHarmonics

paganol:
This PR adresses PR #446. 
In detail:
- Added support for arithmetic operations:
  - `+`, `-` between `SphericalHarmonics` objects
  - `*` with scalar or Stokes vector
- Implemented `.convolve()` method for scalar or per-Stokes ℓ-dependent filters
- Added `.allclose()` for tolerance-based comparison
- Improved Healpy-compatible I/O:
  - `.write_fits()` and `.read_fits()`
- Added utility methods:
  - `lmax_from_num_of_alm()` to infer `lmax` from alm size
  - `get_l_array()` to compute ℓ values in Healpy ordering
- Extended test coverage and updated docstring

### DIFF
--- a/litebird_sim/spherical_harmonics.py
+++ b/litebird_sim/spherical_harmonics.py
@@ -1,8 +1,9 @@
 # -*- encoding: utf-8 -*-
 from dataclasses import dataclass, field
-from typing import Optional
+from typing import Optional, Union
 
 import numpy as np
+import healpy as hp
 
 
 @dataclass
@@ -18,6 +19,7 @@ class SphericalHarmonics:
     This small data class keeps the array and the values `l_max` and `m_max` together.
     The shape of `values` is *always* ``(nstokes, ncoeff)``, even if ``nstokes == 1``.
 
+    It also provides algebraic operations and I/O utilities compatible with Healpy.
 
     Attributes
     ----------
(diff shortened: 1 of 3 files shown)


litebird_sim/spherical_harmonics.py
```diff
@@ -221,3 +316,213 @@ def resize_alm(
...
+
+    def __add__(self, other):
+        if not isinstance(other, SphericalHarmonics):
+            raise TypeError("Can only add another SphericalHarmonics object")
+
+        if (
+            self.lmax != other.lmax
+            or self.mmax != other.mmax
+            or self.nstokes != other.nstokes
+        ):
```
ziotom78, line 332: This test occurs several times. Maybe wrapping it in a function `is_consistent(self, other: SphericalHarmonics) -> bool` would make the code shorter and more readable?

litebird_sim/spherical_harmonics.py
```diff
@@ -221,3 +316,213 @@ def resize_alm(
...
+            raise ValueError(
+                "SphericalHarmonics objects must have matching lmax, mmax, and nstokes"
+            )
+        self.values -= other.values
+        return self
+
+    def __eq__(self, other):
+        if not isinstance(other, SphericalHarmonics):
+            return False
```
ziotom78, line 452: This condition means that the user is comparing apples with oranges, which is probably not what they want. Perhaps it would be better to raise a `ValueError` exception, stating that only `SphericalHarmonics` objects can be compared.

litebird_sim/spherical_harmonics.py
```diff
@@ -221,3 +316,213 @@ def resize_alm(
...
+            self.lmax == other.lmax
+            and self.mmax == other.mmax
+            and self.nstokes == other.nstokes
+            and np.array_equal(self.values, other.values)
+        )
+
+    def allclose(self, other, rtol=1e-5, atol=1e-8):
+        """Compares SH values with tolerance."""
+        if not isinstance(other, SphericalHarmonics):
+            return False
```
ziotom78, line 463: Same here

ziotom78 (commented):
It looks fine to me! I added just a few minor comments.

PR #450, opened 2025-07-05
Add Interface Hierarchy page to documentation

paganol:
This PR adds a new documentation page describing the interface hierarchy.
We could use it as an opportunity to review and improve other parts of the documentation as well, if you think they need it.


ziotom78:
Tests are failing because they're taking too much time: The five slowest tests take ~2 minutes to run, which is way too much. We will trim these tests later; for the moment, I am merging this.

Issue #455, opened 2025-08-04
Incorrect Handling of HWP with Precomputed Pointing


paganol:
In the following modules:
- scan_map.py
- beam_convolution.py
- Map-makers (e.g., binner, destriper, GLS)

When using precomputed pointing, the framework silently ignores any HWP modulation if the hwp argument is None. This happens in several parts of the framework that rely on the following conditional pattern:
```
if hwp is None:
    hwp_angle = None
else:
    hwp_angle = _get_hwp_angle(
        obs=cur_obs, hwp=hwp, pointing_dtype=pointings_dtype
    )
```

In all the modules listed above, when the pointing is pre-computed the framework now follows this logic:
- `_normalize_observations_and_pointings` returns the `obs.pointing_matrix`
- This gets passed to the low level function and then to `_get_pointings_array` that, if hwp_angle is `None`, returns `None` even if the observation has the hwp_angle precomputed 

Thus the HWP algebra is never applied in these modules when using precomputed pointing and hwp=None.

ziotom78:
For the record, it appears that #444 is the cause of this issue.

PR #456, opened 2025-08-04
Fix hwp_angle with precomputed pointing

paganol:
This PR addresses the issue #455. 
Now in `scan_map.py`, `beam_convolution.py`, and the map-makers, the `hwp_angle` is always computed or retrieved inside observation-level functions (e.g., `scan_map_in_observations`), even when hwp=`None`.
This fixes the inconsistency in the algebra adopted when we precompute the pointings.


ziotom78:
@martamonelli, could you please share the code that led you to discover the problem? It would be best to add a new test case where we switch the HWP and verify that the bug won’t reappear.

martamonelli:
@ziotom78, this is how I was setting the different HWPs to simulate the TOD and make the maps. I cannot attach the full notebook to this comment, so I've sent it via email in case it's useful.

```
ang_speed_radpsec = 46 * 2 * np.pi / 60
ang_speed_DIFF_amplitude_radpsec = ang_speed_radpsec/100
ang_speed_DIFF_ang_speed_radpsec = 4.4
ang_speed_DIFF_start_angle_rad = 0

#rotating HWP
sim.set_hwp(hwp=lbs.IdealHWP(ang_speed_radpsec))

sim.prepare_pointings()
sim.precompute_pointings()

sim.nullify_tod()
sim.fill_tods()

tod_rot = np.copy(sim.observations[0].tod[0])

####

start_angle_rad = np.deg2rad(5)

#rotating HWP with miscalibration
sim.set_hwp(hwp=lbs.IdealHWP(ang_speed_radpsec,start_angle_rad=start_angle_rad))

sim.prepare_pointings()
sim.precompute_pointings()

binner_results = sim.make_binned_map(
    nside=nside,  # one can set also a different resolution than the input map
    output_coordinate_system=lbs.CoordinateSystem.Galactic,
)

m_cmb_fg = binner_results.binned_map  # this is the map coming from the binned map-maker
```

ziotom78:
I fixed the conflicts, now all the tests pass. Can we merge this, or is there another step required?

paganol:
I think we can merge it

PR #463, opened 2025-09-22
New proposition for HWP handling

mj-gomes:
_This PR is an alternative to https://github.com/litebird/litebird_sim/pull/454, which I think is cleaner. If this one is to be preferred, the other one will be obsolete._

Hi, as I was finishing https://github.com/litebird/litebird_sim/pull/454, I realized that it was still not the cleanest way to deal with HWP in LBS. Namely:

1. we still have a hwp abstract class that was used only for declaring a ideal case hwp, and even when we are using a non-ideal HWP, we still create a IdealHWP object, which is confusing.
2. The whole hwp_sys class was confusing because we can have a case with a non ideal hwp computed with scan_map; what we call "hwp_sys" should be "hwp_harmonics", because that is the only case where it is really different from scan_map.
3. In the version I was developing in the other branch, the need to specify what kind of hwp matrices we were dealing with only in scan_map_in_observations seemed confusing, as we still have the problem in point 1, mixing Ideal with Non Ideal HWP for the same scenario.

I decided to create a new class, inherited from HWP, which is NonIdealHWP, and can take two additional arguments, which will describe the type of matrices we will deal with: if it's jones or mueller, and if we are considering rotation harmonics expansion or not.

image

This is done at the very beginning of our scripts, when we create our hwp object, and then we don't have to worry about it "anymore": **scan_map_in_observations will know what to do**. We just have to provide the parameters that we used to provide to hwp_sys.fill_tod when we are dealing with harmonics.

This makes much more sense to me, as when we write a script, we only have to deal with choosing the type and formalism for the HWP when creating the HWP object itself, which seems intuitive and less prone to confusion and errors.

There are several details that I will explain better in the next call, as it would make a huge text in here. Does this seem like a good idea for you?


ziotom78:
Hi Miguel, thanks! I have added this item in the agenda for the next ST telecon.

(For the moment, I have a quick suggestion: instead of using a string to encode the “formalism”, it would be better to have an enumeration: otherwise, people might wonder whether the parameter should be `jones` or `Jones`.)

mj-gomes:
> Hi Miguel, thanks! I have added this item in the agenda for the next ST telecon.
> 
> (For the moment, I have a quick suggestion: instead of using a string to encode the “formalism”, it would be better to have an enumeration: otherwise, people might wonder whether the parameter should be `jones` or `Jones`.)

Do you mean an enumeration as in an Enum type class where I define the different ways to write jones/mueller (capitalized or not), associating them with a single value, or do you mean simply enumerate the options by using integers e.g. 1 for jones and 2 for mueller as arguments instead of strings ?

If it is the first option, isn't it the same as just setting a line in the initialization of the class like
`if formalism in ['jones', 'Jones', 'JONES']` ?

Or do you mean something else?

mj-gomes:
@ziotom78, I agree that  Enum works like a charm for PyCharm (that must be why it is called PyCharm in the first place).

The only thing I am "afraid" if we use only Enum is that it becomes less intuitive for people not using PyCharm or similar, as they will have to go search for the options and type **Formalism.MUELLER**, which is less smooth of a process then simply writing a string. Then again, I would say the majority of users will be using VS Code or PyCharm or Spyder, and all of these automatically list the possible values, so it is probably worth the small risk of some people having to read the function description, and we get the benefits you mentioned.

Otherwise, we could also simply do it as it was done before in HwpSys, naming the argument **mueller_or_jones**, which should be pretty self-explanatory of the two options and their lower case.

mj-gomes:
I will apply the Enum option, thank you for the suggestion and explanation !

Should we do this for other similar situations in lbs to have the type check flagging of incorrect values everywhere? I am thinking for example in "input_map_in_galactic" in scan_map, which could become an Enum "Coordinates"; or even the pointings_dtype, but that one would be a change everywhere in lbs, maybe the confusion for everyone used to lbs makes it not worth changing. What do you think ?

PR #465, opened 2025-09-30
Add and update actions

ggalloni:
This PR adds a few things for the actions:

- Dependabot config
- update uv.lock in PRs from Dependabot
- automatic PyPI publishing + email text generation

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -6,6 +6,11 @@ jobs:
   tests:
     runs-on: ubuntu-latest
 
+    # Only run tests if there is no [skipci] or [skip ci] in the commit message
+    if: |
+      !startsWith(github.event.head_commit.message, '[skip ci]') &&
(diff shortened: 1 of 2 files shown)


ggalloni:
@ziotom78, simply updating `macos-13` to `macos-latest` (`macos-15`) improved quite a bit the performance. Also, `macos-13` is deprecated from like today, so we would have moved to something more recent anyway...
Now, `ubuntu-latest` and `macos-latest` builds perform similarly. 
Still, it look like `macos-latest` struggles a bit to find workers, so maybe we should test `macos-14`, what do you think?

`Code Style` is now tested separately :+1: 

I did the same for `Docs Tests`. 
I don't think we need to run that for every combination of OS and Python version, right?

Also, there are a couple of `appveyor_*` scripts and a hook folder, which are seemingly unused. Is it OK to remove them, or am I missing something? 

The `run_tests.sh`, `run_tests.bat`, and `refresh_docs.sh` scripts are just convenience scripts at this point, so I guess that we could keep them.

PR #473, opened 2025-12-04
Implementation of h_n maps

camilledemay:
WIP: The goal is to have a module to generates the H_n maps described in https://arxiv.org/abs/2109.05038


litebird_sim/mapmaking/h_n.py
```diff
@@ -0,0 +1,370 @@
...
+@dataclass
+class HnMapResult:
+    """Result of a call to the :func:`make_hn_maps` function
+
+    This dataclass has the following fields:
+
+    - ``h_maps``: Dictionnary containing the h_n maps for each spin order n,m and each detector.
+
+    - ``coordinate_system``: the coordinate system of the output maps
+      (a :class:`.CoordinateSistem` object)
```
Magwos: Little typo (but I see that it is also in the binner object here: https://github.com/litebird/litebird_sim/blob/bdd1fc2b86b1a085284d8655ff4c10f7e6f317f1/litebird_sim/mapmaking/binner.py#L53
```suggestion
      (a :class:`.CoordinateSystem` object)
```

litebird_sim/mapmaking/h_n.py
```diff
@@ -0,0 +1,370 @@
...
+    nobs_matrix: npt.ArrayLike,
+    num_of_detectors: int,
+) -> None:
+    assert pix.shape == psi.shape
+
+    for idet in range(num_of_detectors):
+        if not d_mask[idet]:
+            continue
+
+        print(np.shape(pix[idet]), "   ", np.shape(hwp_angle))
```
Magwos: Delete print? (If print to be added to the logs, maybe it would be useful to add as well the detector id
```suggestion
```

Magwos (commented):
A very few minor comments for the PR + one additional (although probably not necessary) -> Different conventions are used in the comments, either `Args` or `Parameters`, should it be homogenized? (I can propose changes to do so if yes!)

litebird_sim/mapmaking/h_maps.py
```diff
@@ -0,0 +1,418 @@
+# WIP
+# -------
+# -------
```
ziotom78: Is this still a WIP?

litebird_sim/mapmaking/h_maps.py
```diff
@@ -0,0 +1,418 @@
...
+    #
+    # This method alters the parameter `nobs_matrix`, so that after its completion
+    # each 3×3 matrix in nobs_matrix[idx, :, :] will be the *inverse*.
+
+    # Expected shape:
+    # - `nobs_matrix`: array of shape (Ndet,N_p,3), where
+    #   N_p is the number of pixels in the map and Ndet the number of detectors
+    # - `atd`: (Ndet,N_p, 2)
+    npix = atd.shape[0]
+    for ipix in range(npix):
```
ziotom78: Might `prange` here and `@njit(parallel=True)` above improve the performance?

ziotom78 (approved):
Hi @camilledemay, this PR looks fine to me! I just left a couple of minor comments, but once you answer them, you can merge it. (Be sure to update the `CHANGELOG` first!)

litebird_sim/mapmaking/h_maps.py
camilledemay: Hi Maurizio,
No, it's no longer WIP, I'll remove this

litebird_sim/mapmaking/h_maps.py
camilledemay: I just tried this, indeed it is faster with the parallelized for loop

PR #481, opened 2026-01-07
Add type checking with `ty`

ggalloni:
This PR attempts adding `ty` to perform automatic type-checking.

Right now, this the situation in terms of errors detected by `ty` [rule: number of violations] (-> remaining after edit):

- [x] not-subscriptable: 537
- [x] invalid-argument-type: 309
- [x] invalid-assignment: 256
- [x] possibly-missing-attribute: 167
- [x] unresolved-attribute: 102
- [x] unsupported-operator: 60
- [x] invalid-parameter-default: 30
- [x] not-iterable: 25
- [x] unresolved-import: 16
- [x] no-matching-overload: 12
- [x] invalid-return-type: 11
- [x] unresolved-reference: 10
- [x] missing-argument: 8
- [x] parameter-already-assigned: 5
- [x] deprecated: 4
- [x] invalid-type-arguments:  4
- [x] unknown-argument: 3

Thus, this will require some time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # HEAD
 
+-   Add automatic type-checking with ty [#481](https://github.com/litebird/litebird_sim/pull/481)
+
 -   Stop using sphinxcontrib-asciinema [#478](https://github.com/litebird/litebird_sim/pull/478)
 
 -   TODs handling improved [#472](https://github.com/litebird/litebird_sim/pull/472)


ziotom78:
Giacomo, thank you so much for having undertook this ordeal! I know this must have been tedious work, especially considering the broad extent of the changes. It’s impressive to see how many inconsistencies you’ve managed to fix.

I’m particularly pleased to see that `ty` caught auctual bugs, often in seldomly-run code paths. Seeing it propagate architectural changes like the new `SphericalHarmonics` class really reveals how useful it is for a large codebase like LBS. And it’s great that your modifications improve the documentation and adds robustness via several `asserts`. I believe this shows that `ty` effectively enforces code clarity.

Regarding the specific changes:

-   I noticed we used `npt.ArrayLike` far too often when `npt.NDArray` was more appropriate. (It seems I am one of the heaviest offenders here…) Thanks for fixing so many inconsistencies.

-   Correcting cases where types did not match (like `list` vs. non `list`) or parameters were missing `Optional` or `None` despite having a `None` default is important: even if these hints do not change execution, they improve our documentation tremendously. I als agree with your choice to ignore `unresolved-import` and the Numba issues: it’s better to be pragmatic here. Excluding `docs/`, `benchmarks/`, and the legacy `mbs/` folder also makes perfect sense.

-   I was happily surprised that `ty` caught instances of overwriting constants (like `lbs.PTEP_IMO_LOCATION`). It’s also great to see it can handle Jupyter notebooks!

I have one technical curiosity: in several places, you switched to `getattr` instead of using the dot, e.g., `getattr(mpi4py, "__version__")` instead of `mpi4py.__version__`. Do you know why `ty` was complaining about the latter?

Looking ahead, this PR makes me think of two aspects of our workflow:

- You anticipated my concern about CI efficiency! I agree that the `pre-commit` hook might be too strict if it blocks every single commit while someone is still in “creative mode”. Let’s try it for a while, but if becomes a bottleneck, we can consider moving the type check exclusively to the CI level.

- BTW, Since it can be frustrating to run full type checks on every single commit pushed to a PR (sometimes you really want to commit stuff that is not 100% clean year), perhaps we should consider running these checks only when a PR is being merged. While the `[skipci]` trick in the commit message can be used here, not everyone is familiar with it. Moreover, it would save a huge number of CPU time, which is important because in our free plan the limit is not very high and was reached sometimes.

- I see that Astral is releasing new versions of `ty` almost weekly. I would abstain from updating `ty` every time a new version is released; we could just bump the version if we stumble upon a bug, or if it is part of a larger update.

Excellent work. On my side, this should be merged ASAP, so that people can rebase their copy on this.

PR #494, opened 2026-02-10
Fix beam normalization

paganol:
This pull request updates the polarization normalization in the Gaussian beam synthesis routine to match the convention used in the `gauss_bl()` function. As a result, the expected values in the corresponding unit tests have been updated to reflect the new normalization.

Beam synthesis normalization changes:

* Added Challinor spin-2 normalization factor (`exp(2*sigma^2)`) for polarization modes in `gauss_beam_to_alm` to match the normalization used in `gauss_bl()`.

Test updates for new normalization:

* Updated expected values in `test_gaussbeam_circular` in `test/test_beam_synthesis.py` to reflect the new polarization normalization.
* Updated expected values in `test_gaussbeam_elliptical` in `test/test_beam_synthesis.py` for consistency with the new normalization.


paganol:
Now the difference input VS convolved goes down to $10^{-19}$ K in polarization if we set `epsilon=1e-13` in `BeamConvolutionParameters`. With the default, i.e. `epsilon=1e-5`, we reach $10^{-11}$ K. updated version of the notebook 
[Here](https://github.com/user-attachments/files/25217218/validation_offset_beam.ipynb) an updated version of the notebook

PR #506, opened 2026-03-03
Center pointings

### LINKED ISSUE #504: Possibility of pixel-centering the pointings
I was wondering whether it would make sense to add a function that forces the pointings at the centers of the pixels for a given `nside`. This could be useful to ensure that the pixels used when scanning are exactly the same as the ones used at the map-making step.

I discussed this with @paganol last week, and he suggested introducing a new function, at the same level as `precompute_pointings`, that centers the pointings and reassigns them accordingly.

mj-gomes:
This PR implements the request in https://github.com/litebird/litebird_sim/issues/504.

I implemented the centering by doing ang2pix and pix2ang in sequence, I believe this is the way to do it, as healpix will recompute the angles based on the center of the pixels, but tell me if there is a better way to do this.

p.s. Right now the center_pointings method should always be called after precompute_pointings. I was thinking that maybe we should add the option to center in the other cases, since there is the option to scan a map without precomputing the pointings, and in that case the pointings are computed on the fly and, as of now, not centered. What do you think? Should we add a flag to scan_map_in_observations (and the low level methods that it calls like fill_tod and get_pointings) that will allow to center the pointings also when they are computed on the fly?


ziotom78 (commented):
Hi Miguel. Thanks for being so quick! Luca, Marta, and I had a chat about this, and we think it would be better to support an “on-the-fly” computation of the centered pointings. This has the advantage of making it super-easy to debug cases like Marta's; one can just call `get_pointings(center=True)` to get the centered coordinates for saving or plotting without affecting the rest of the simulation. Usually, we don't want to force centering globally.

It should be quite easy to implement, as LBS already has a private function, `_get_centered_pointings()` (in `pointings_in_obs.py`), which takes a pointing matrix and returns the “re-centered” version. The code is already tested and could be included in the `Observation.get_pointings()` method (in `observations.py`). It would be triggered if `get_pointings()` receives a new Boolean parameter (`center`) set to `True`.

What do you think? Does this approach sound reasonable to you?

ziotom78 (commented):
Hi Miguel, thanks! My suggestion to use `int | None` for `nside_centering` was that this makes `center` redundant: you already know if you want centering or not if `nside_centering` is `None` or an integer. So, `center` can be removed.

ziotom78 (approved):
Thanks for your work, @mj-gomes , it looks ok to me!

paganol:
Hi @mj-gomes @ziotom78, I removed `_get_centered_pointings` function and its centering logic is now handled directly inside `_get_pointings_array`. This should be cleaner and more redable.

If you like this last change, I think we can merge the PR.

paganol (approved):

mj-gomes:
Thanks @paganol, it looks good to me. I merged the master branch and updated the changelog, should be good to go!

PR #511, opened 2026-03-24
fix random number generation in non linearity module

### LINKED ISSUE #510: Seeding problem for mpi division in time
I was running a simulation with the non linearity module, using 4 detectors and 10 MPI tasks, dividing the simulation in time blocks (not in detector blocks). So all of the MPI tasks work on the 4 detectors, but on a different bunch of time for each of them. 

Doing this, I realized that the g_one_over_k value sampled for each detector was different in each MPI task, while it should remain the same (there should be a single value of g for a given detector).

It seems like the seeding process [implemented last year](https://github.com/litebird/litebird_sim/pull/414) assumes that we divide the observation into parallel detector blocks and for now does not work in the case were the division is done in time.

I will look into this, meanwhile if someone has a quick fix, feel free to step in.

mj-gomes:
This PR should fix https://github.com/litebird/litebird_sim/issues/510. 

With these changes, the Non Linearity module does not follow anymore the RNG hierarchy defined in the seeding module, as we need to make sure that the randomly generated g value of the detector non linearity equation is the same for every detector, even if multiple MPI tasks perform operations on the same detector. 

Now, by default, the code takes the name of each detector, generates its ASCII code and uses it as a seed to create an individual PCG64 RNG for each detector, and from there the process is the same as before, i.e. for each detector, we sample from its generator, ensuring the same values are used for all MPI tasks dealing with a given detector, while maintaning reproducibility and the same level of _uncorrelation_ between generated values as before, as we define a new generator for each detector.

There is also still the possibility to overwrite the default seeds from the detector names, by giving a user_seed that will then spawn different generators, one per detector.

I tested it and it seems to be working correctly.

### DIFF
--- a/litebird_sim/non_linearity.py
+++ b/litebird_sim/non_linearity.py
@@ -94,8 +94,6 @@ def apply_quadratic_nonlin_for_one_detector(
         scale=nl_params.sampling_gaussian_scale,
     )
 
-    print("g_one_over_k", g_one_over_k)
-
     for i in range(len(tod_det)):
         tod_det[i] += g_one_over_k * tod_det[i] ** 2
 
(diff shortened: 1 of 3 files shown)


mj-gomes:
I am closing this PR because https://github.com/litebird/litebird_sim/pull/523 will provide a more robust fix for this problem.

PR #519, opened 2026-04-14
Support Python 3.14 and upgrade NumPy to 2.x

ziotom78:
With the release of PySM3 3.4.4, we can finally add support for Python 3.14 and update a few dependencies; most notably, we can now drop support for NumPy 1.x and move to NumPy 2.x.

The transition was not seamless, though. It seems that `np.testing.assert_allclose()` causes weird memory corruption errors when used with Py 3.14 + NumPy 2.2, so I had to change it to `assert np.allclose()` (which does not seem to suffer from the same problem).

Because of an updated dependency on AstroPy, a few tests that relied on HealPy to convert Equatorial coordinates into Galactic coordinates started failing, probably because of an update in the transformation matrices used by AstroPy. In a few cases I modified tests to stop using HealPy to do the conversion, but for `test_scan_map.py` I realized that the very long test on the binner could be significantly simplified, as in the first part it just re-implemented the map-making equation to compare it with the result of the binner. As a consequence, this test now runs in ~50% of the time.


ziotom78:
I forgot to say that, unlike NumPy and AstroPy, we are **not** dropping support for Python 3.10, which is the reason why I did not upgrade NumPy and AstroPy to the latest version.

The End of Life for Python 3.10 is scheduled for October 2026, so we have still some time before dropping support for it.

anand-avinash:
Hi @ziotom78! I have tested this PR with Python v3.14.4 and NumPy v2.2.6. All tests are passing. I also run the notebook examples, it worked without issues, except when plotting the healpix maps with `hp.mollview`. For instance, in [litebird_sim_example.ipynb](https://github.com/litebird/litebird_sim/blob/3a8e72c1de03b68843e6ad4da7e58447c980a276/notebooks/litebird_sim_example.ipynb), it produces several lines of the following error:

```
Error encountered at [/project/cextern/healpix/src/cxx/Healpix_cxx/healpix_base.cc](https://file+.vscode-resource.vscode-cdn.net/project/cextern/healpix/src/cxx/Healpix_cxx/healpix_base.cc), line 888
(function I T_Healpix_Base<I>::loc2pix(double, double, double, bool) const [with I = long int])

must not happen
```

with the following maps:
image

In the current master branch, it works as expected.

ziotom78:
> Hi @ziotom78! I have tested this PR with Python v3.14.4 and NumPy v2.2.6. All tests are passing. I also run the notebook examples, it worked without issues, except when plotting the healpix maps with `hp.mollview`.

Hi @anand-avinash, I tried to replicate your problems, but on my case (Endeavour OS Linux, Python 3.14.5, NumPy 2.2.6, Healpy 1.19.0) everything works as expected:

debug1
debug2
debug3

@ggalloni , @paganol , can you open `notebooks/litebird_sim_example.ipynb` and run it?

mj-gomes:
@anand-avinash were your tests done on a windows machine? There were some default integer dtype changes in windows in numpy 2 that could eventually point us to the reason behind this (see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#windows-default-integer)

anand-avinash:
Seems very strange! I tested it again in a new environment and it produced the same error. I am running the code on my laptop with Almalinux 10.1 with the following steps for setup.

```bash
git clone https://github.com/litebird/litebird_sim.git lbsim_healpy
cd lbsim_healpy
git checkout numpy2_py314
conda create -n lbsim_healpy python=3.14.4
conda activate lbsim_healpy
pip install numpy==2.2.6
pip install -e .
pip install ipykernel
```
With these, I get healpy v1.19.0.

ziotom78:
I see that you are using Python 3.14.4, while I am using 3.14.5. Might be that the new bugfix release fixed that very issue?

According to the [release notes](https://www.python.org/downloads/release/python-3145/), they reverted the garbage collector back to what Python 3.13 used (plus “154 bugfixes”).

paganol:
Hi all, I tested this branch on G100 using the new version of PySM, [3.4.5](https://github.com/galsci/pysm/releases/tag/3.4.5), and everything now works smoothly.
I think we can update `pyproject.toml`, merge this PR and probably take the opportunity to release version 0.17.0.

PR #524, opened 2026-05-22
Add possibility to convert Non Linearity factor from 1/K to 1/SR

mj-gomes:
This is one other feature that I had been using in the HWP simulations and that was not yet in lbs.

It allows to pass a flag to `sim.apply_quadratic_nonlin` to convert the non linearity factor (g) from 1/K to 1/SR (spectral radiance units), so that we keep coherent units when computing the TOD in spectral radiance units and then inserting the non linearity effect on it.

Should we also make this information (the conversion flag and also the sampled value for the non linearity factor) available in the automatic output simulation report?


ziotom78:
Hi @mj-gomes, these changes look fine to me.

I believe that this information should definitely be included in the report! In this way, whoever reads it can check if the units were used consistently.

I would add an item about `convert_K_to_SR` in the docstring for `Simulation.apply_quadratic_nonlin` (just copy the one from `apply_quadratic_nonlin_for_one_detector`).

Thanks!

mj-gomes:
> Hi @mj-gomes, these changes look fine to me.
> 
> I believe that this information should definitely be included in the report! In this way, whoever reads it can check if the units were used consistently.
> 
> I would add an item about `convert_K_to_SR` in the docstring for `Simulation.apply_quadratic_nonlin` (just copy the one from `apply_quadratic_nonlin_for_one_detector`).
> 
> Thanks!

It's done. It's the first time I deal with the append_to_report method, but from what I understand it is sufficient adding conv_K_to_SR as an argument, since the method deals automatically with the kwargs. Is this correct?

If so, I would say this can be merged (in any case it does not introduce breaking changes to anything already implemented).

ziotom78:
> It's done. It's the first time I deal with the append_to_report method, but from what I understand it is sufficient adding conv_K_to_SR as an argument, since the method deals automatically with the kwargs. Is this correct?

Partially, but this is not enough to make the value of the parameter appear in the report. Internally, LBS uses the [Jinja2](https://jinja.palletsprojects.com/en/stable/) template engine to produce the Markdown reports, so you must make sure that the new variable appears somewhere in the final report.

Take a look at [litebird_sim/templates/report_mpi.md](https://raw.githubusercontent.com/litebird/litebird_sim/refs/heads/master/litebird_sim/templates/report_mpi.md): it uses the variables `mpi4py_version`, `mpi_implementation`, `mpi_version`, and `warning_mpi_version` to decide what to put in the report. In your case, I believe that the best option would be to use an [`if`](https://jinja.palletsprojects.com/en/stable/templates/#if). Something like

```
{% if conv_K_to_SR -%}
Put the text that should be produced when the variable is True
{% else %}
Put the text that should be produced when the variable is False
{% endif %}
```

BTW, I couldn’t find any template report for the HWP non-linearity module in the [templates](https://github.com/litebird/litebird_sim/tree/master/litebird_sim/templates) folder, so you should probably create one first.

mj-gomes:
Hi @ziotom78 it is done. I tested on my side and it is outputting everything to the report. It is outputting the distribution parameters from the non linearity factor sampling, which should be enough given that the exact values sampled are reproducible with the same distribution. Maybe we could later make the exact sampled g values be on the report, but since that means getting the values for every detector, and thus communicate them between different MPI tasks, I did not do it for now, just to make this faster to merge so that I can have a solid lbs version to refer to in my simulations. If you agree, and if we could merge this today, I would have almost everything needed for the HWP simulatons in the main branch (the only thing missing is the one in Avinash's PR, about the seeding). Thanks !

ziotom78:
That’s fine, @mj-gomes, the report is good enough for the moment. Please merge it!

PR #549, opened 2026-08-06
Fix input sky unit error

### LINKED ISSUE #548: Unit problem with bandpass_integration in input sky generation
It seems that when bandpass_integration is set to True in SkyGenerationParams, the output units of the generated maps are incorrect. 

Comparing a map generated for the same channel, using only the center frequency vs integrating over the band, we have a factor 1e6 between them. It happens with uK_CMB and K_CMB, so it doesn't seem like a simple conversion missing from one to the other.

I'll look into the source problem as soon as I can. Meanwhile, here is a simple script showing the problem.

_**UPDATE:** I just realized that this happens for the CMB but not for the foregrounds, which may make it easier to find the problem._

```
imo = lbs.Imo(flatfile_location="schema.json")

chinfo = lbs.FreqChannelInfo.from_imo(
     url=f"/releases/IMo_vPostKDP2_Option2/LMHFT/L1_L/channel_info",
     imo=imo,
)
    
params = SkyGenerationParams(
        nside=256,
        bandpass_integration=True,
        units=Units.uK_CMB,
        output_type="map",
        make_cmb=True,
        seed_cmb=0,
        make_fg=False,
        make_dipole=False,
        apply_beam=True,
    )

sky_gen = SkyGenerator(
            parameters=params,
            channels=chinfo,
        )
sky_cmb = sky_gen.execute()

params = SkyGenerationParams(
        nside=256,
        bandpass_integration=False,
        units=Units.uK_CMB,
        output_type="map",
        make_cmb=True,
        seed_cmb=0,
        make_fg=False,
        make_dipole=False,
        apply_beam=True,
    )

sky_gen = SkyGenerator(
            parameters=params,
            channels=chinfo,
        )
sky_cmb_noBP = sky_gen.execute()

plt.figure()
for i in range(3):
      hp.mollview(sky_cmb["L1_L"].values[i], sub=(2,3,i+1))
      hp.mollview(sky_cmb_noBP["L1_L"].values[i], sub=(2,3,i+4))

plt.show()

```

Here is the output:

Image

mj-gomes:
This PR fixes #548 

`origin_unit` in [_get_cmb_unit_conversion()](https://github.com/litebird/litebird_sim/blob/30d61720baf5b479c0b2db6b8e12ca59305288c8/litebird_sim/input_sky.py#L32) is by default K_CMB, but then when calling [pysm3.bandpass_unit_conversion](https://github.com/litebird/litebird_sim/blob/30d61720baf5b479c0b2db6b8e12ca59305288c8/litebird_sim/input_sky.py#L102) we do not specify that as the input unit, so it was using the default uK_RJ.

Now it is working.

~~TODO before merging:~~ (DONE)

- [x] Fix expected values in tests


mj-gomes:
Added a CI test for the verification of correct conversion when bandpass integrating.

mj-gomes:
I noticed something which can be improved for tidiness. Currently, we assume the input cmb power spectrum used to generate the cmb map is always in uK_CMB (see [here](https://github.com/litebird/litebird_sim/blob/30d61720baf5b479c0b2db6b8e12ca59305288c8/litebird_sim/input_sky.py#L168)), and then we apply the scaling factor so that our base units are K_CMB. We then perform conversions from that on, and that is coherent with the default unit in [here](https://github.com/litebird/litebird_sim/blob/7e168323c1eee4bf8371e656268505562fdc5acc/litebird_sim/input_sky.py#L32) being K_CMB. 

For this reason, the following part of the cmb unit conversion code is never reached:

https://github.com/litebird/litebird_sim/blob/7e168323c1eee4bf8371e656268505562fdc5acc/litebird_sim/input_sky.py#L109-L118


We should either delete this part, or relax the assumption that the input power spectrum is always in uK_CMB (and in that case add a parameter to SkyGenerator stating the unit of the input spectrum in case one is given).

What do you think? I suppose we always use uK_CMB, so I don't know if the second option is relevant, but in that case we can just delete the code.

paganol:
Pushed c69a6690 on top of this branch to resolve the `origin_unit` question raised above.

Answer: delete it, and it's not just "unreachable in practice" — it's redundant regardless of `origin_unit`. `pysm3.bandpass_unit_conversion(freqs, weights, output_unit, input_unit)` already computes the full `input_unit -> output_unit` factor directly. Once this PR's fix passes `origin_astropy` as `input_unit`, that single call already gives the correct `origin_unit -> target_unit` factor for *any* origin unit — so the old `if origin_unit == K_CMB: return ... else: compute factor_to_origin and take the ratio` block was a workaround for not being able to pass `input_unit` at all, and is now dead code even in the hypothetical case where `origin_unit != K_CMB`.

```python
factor_to_target = pysm3.bandpass_unit_conversion(
    freqs, weights, target_astropy, origin_astropy
)
return factor_to_target.value
```

Checked:
- `test/test_input_sky.py` (all 11 tests, including your new `test_bandpass_integration_units_basic`) still passes.
- The dipole path goes through this same function (`_cmb_unit_conversion_for_target` is shared between CMB and dipole generation), so this fix also resolves the same bandpass-integration unit bug for the dipole. Confirmed empirically: with `make_dipole=True`, `bandpass_integration=True` vs `False` now match to ~1e-12, same as CMB.

paganol (approved):

paganol:
Hi @mj-gomes, feel free to merge the code whenever you like

PR #550, opened 2026-08-25
Homogenize nthreads handling across the framework

### LINKED ISSUE #544: Missing `nthreads` arguments and its inconsistent defaults
At some places in the code, the argument for `nthreads` in relevant `ducc` functions are still missing. While it is quickly fixable, I noticed that its default values are set inconsistently. At some places the default is set with `nthreads: int | None = None` usually followed by a call `nthreads = resolve_nthreads(nthreads)`, while at others the default is set with `nthreads: int = 0`. Also in `resolve_threads()`, when `nthreads` is `None`, it returns `0`. With `0` default, `ducc` would use all available hardware threads. With multiple MPI processes, this may lead to thread contention and may affect the performance.

As for consistency in setting the defaults,I was curious what could be the correct way to resolve this. One would be to use consistent defaults in all cases, followed by the `resolve_nthreads()` call. Another solution would be to make `nthreads` a global variable in `litebird_sim` (much like `MPI_COMM_WORLD`) and use it consistently without having to call `resolve_nthreads()` every time `nthreads` is needed.

Please let me know what you think.

paganol:
Fixes #544.

Standardizes every `nthreads` parameter on `int | None = None` + `resolve_nthreads()`, removing the previous mix of `0`/`1` literal defaults and duplicated `OMP_NUM_THREADS`-reading code that motivated the issue.

## Summary

- `utilities.py`: `NUM_THREADS`/`NUMBA_NUM_THREADS` are now resolved **once at import time** (mirrors `lbs.MPI_COMM_WORLD`) instead of re-reading the environment on every call. `resolve_nthreads()`/`resolve_numba_nthreads()` become cheap lookups against the cached value, and both globals are exposed at the top-level `lbs` namespace.
- `constants.py`: `NUMBA_NUM_THREADS_ENVVAR` was accidentally aliased to `"OMP_NUM_THREADS"` (identical to `NUM_THREADS_ENVVAR`) — fixed to point at its own `NUMBA_NUM_THREADS` variable, so Numba and ducc0 thread counts can be tuned independently, with `OMP_NUM_THREADS` as the shared fallback when `NUMBA_NUM_THREADS` isn't set.
- `simulations.py`: `Simulation.numba_threads` now resolves via `resolve_numba_nthreads()`, called *after* `_init_missing_params()` so a `numba_threads` value from a parameter file still takes precedence over the environment (precedence: explicit constructor arg > parameter file > `NUMBA_NUM_THREADS` > `OMP_NUM_THREADS` > hardware default).
- `hwp_harmonics.py`: replaced an inline reimplementation of `resolve_nthreads()`'s logic with the shared helper.
- `beam_convolution.py`, `maps_and_harmonics.py` (`interpolate_alm`, `pixelize_alm`, `estimate_alm`, `rotate_alm`), `scan_map.py`, `pointings_in_obs.py`, `input_sky.py`: replaced literal `nthreads=0` defaults with `None` + `resolve_nthreads()`.
- Also threaded already-resolved `nthreads` values into internal `_get_pointings_array()` calls in `beam_convolution.py`, `scan_map.py`, `hwp_harmonics.py`, and `mapmaking/common.py` that had a resolved value in scope but weren't passing it through, silently falling back to ducc0's own default instead.
- `grasp2alm.py`: `PseudoAnalysisSHT.to_alm()` gained an `nthreads` parameter (was hardcoded to `0` in two direct ducc0 calls, with no way to override).
- `mueller_convolver.py` intentionally left untouched: it's vendored from upstream ("maintained externally... to make updating from outside sources easier") and its only call site already passes an explicit resolved value.

## Testing

- Full test suite: 277 passed, 2 skipped (non-MPI) + `test_mpi_n4.py` passed.
- Manually verified env-var precedence (`OMP_NUM_THREADS` only, `OMP_NUM_THREADS` + `NUMBA_NUM_THREADS` together) and `Simulation` precedence (explicit arg > parameter-file value > env > hardware default).
- `ruff check` clean.

## Test plan

- [x] `pytest test/` (non-MPI)
- [x] `pytest test/test_mpi_n4.py`
- [x] Manual precedence checks for `resolve_nthreads`/`resolve_numba_nthreads`
- [x] Manual precedence check for `Simulation(numba_threads=...)` vs parameter-file `numba_threads` vs env vars


litebird_sim/input_sky.py
anand-avinash: `nthreads` arg is missing here

litebird_sim/mapmaking/common.py
anand-avinash: `nthreads` arg is missing here

litebird_sim/utilities.py
```diff
@@ -1,16 +1,60 @@
...
+def _hardware_threads() -> int:
+    """Number of threads available to this process (affinity-aware on Linux)."""
+
+    return ducc0.misc.available_hardware_threads()
+
+
+def _compute_nthreads() -> int:
+    if NUM_THREADS_ENVVAR in os.environ:
+        return int(os.environ[NUM_THREADS_ENVVAR])
+    return _hardware_threads()
```
anand-avinash: This will lead to using all available processes for threading, causing contention. Would be better if we simply set it to return 1, when no envvar is set.

anand-avinash (commented):
Everything looks good except few missed threading opportunity and addressing the contention.

litebird_sim/utilities.py
paganol: Hi @anand-avinash, I made some test here and it seems to work fine on clusters. Can we keep as it is and invest more on the documentation? misc.available_hardware_threads of ducc0 seems to work very well. I've also checked whether numba/ducc0 could conflict directly, but I could not find any case.

